PR #12, opened 2019-06-09
Static types for ERTP with flow (WIP)

dckc:
migrated from https://github.com/Agoric/SwingSet/pull/65


erights:
@dckc @michaelfig can we close this as obsolete? If we did revisit this, I suspect we'd start again with lessons from this one, rather than revising this one?

dckc:
yes; I don't use this and haven't maintained it.

PR #19, opened 2019-06-14
more/pixels/gallery: fake out randomness

michaelfig:
Make the initial pixels at least a little more colourful.  More can be done.


michaelfig:
@katelynsills tests are failing, but I don't think that's caused by my change.

katelynsills:
> @katelynsills tests are failing, but I don't think that's caused by my change.

I found the test, it was checking that the color changed appropriately and had a log of the original color, which changed

PR #31, opened 2019-06-20
Test handoff

Chris-Hibbert:
A test to demonstrate Alice selling pixels to Bob through the handoff service.

### DIFF
--- a/demo/gallery/bootstrap.js
+++ b/demo/gallery/bootstrap.js
@@ -42,7 +42,7 @@ function build(E, log) {
     const aliceP = E(aliceMaker).make(userFacet);
     const bobP = E(bobMaker).make(userFacet);
     const { dustIssuer } = userFacet.getIssuers();
-    const aliceDustPurseP = dustIssuer.makeEmptyPurse('Alice dust purse');
+    const aliceDust = dustIssuer.makeEmptyPurse('Alice dust purse');
     const bobDust = adminFacet.dustMint.mint(37, 'bob purse');
     const {
       aliceRefundP,
@@ -51,7 +51,7 @@ function build(E, log) {
       contractHostReceipt,
     } = await E(aliceP).doTapFaucetAndOfferViaCorkboard(
       handoff,
-      aliceDustPurseP,
+      aliceDust,
     );
     // Don't start Bob until Alice has created and stored the buyerSeat
     const { bobRefundP, bobPixelP } = await Promise.all([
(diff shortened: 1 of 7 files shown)


demo/gallery/bootstrap.js
```diff
@@ -36,6 +36,33 @@ function build(E, log) {
...
     await E(aliceP).doSellAndBuy();
   }
+  async function testAliceSellsToBob(aliceMaker, bobMaker, gallery, handoff) {
+    log('starting testAliceSellsToBob');
+    const { userFacet, adminFacet } = gallery;
+    const aliceP = E(aliceMaker).make(userFacet);
+    const bobP = E(bobMaker).make(userFacet);
+    const { dustIssuer } = userFacet.getIssuers();
+    const aliceDustPurseP = dustIssuer.makeEmptyPurse('Alice dust purse');
+    const bobDust = adminFacet.dustMint.mint(37, 'bob purse');
```
katelynsills: I think that both alice's dust purse and bob's dust purse are the same type, and (I could be wrong on this) but neither are presences or promises.

demo/gallery/vat-bob.js
```diff
@@ -52,6 +53,44 @@ function makeBobMaker(E, log) {
...
+          const dustRefundP = E(dustIssuer).makeEmptyPurse('dust refund');
+          await collect(buyerSeatP, pixelPurseP, dustRefundP, 'bob option');
+
+          const exclusivePayment = await E(pixelIssuer).getExclusiveAll(
+            pixelPurseP,
+          );
+
+          const { useRightPayment } = await E(gallery).split(exclusivePayment);
+
+          const useRightIssuer = E(useRightPayment).getIssuer();
```
katelynsills: I think we should use the useRightIssuer from the gallery separately rather than getting it from the useRightPayment. In both cases, we're trusting the gallery, but getting the issuer separately allows us to do a double check

demo/gallery/bootstrap.js
Chris-Hibbert: I think you're right that the gallery, and thus the issuers are local to the bootstrap vat. 

Are you suggesting that since they're from the same issuer, the same facet could make both purses? makeEmptyPurse() and mint() are only available from separate facets. Would it be better to use mint(0, ...) to make Alice's purse?

katelynsills: Oh, I was talking about the 'P' suffix and just saying that it seems like they should either both have it or not have it. I think in this case there shouldn't be a P suffix.

demo/gallery/vat-bob.js
Chris-Hibbert: done.

PR #33, opened 2019-07-08
Refactor verification

Chris-Hibbert:
refactor contractHost to simplify verification of invites.  …
Also add a helper for building amounts.

I wonder whether there's a way to move the added complexity of defining
contracts into the contractHost.

const escrowExchangeThunk = () =>
  harden({
    start(terms, inviteMaker) {

vs.

function escrowExchange(terms, inviteMaker) {

### DIFF
--- a/core/contractHost.js
+++ b/core/contractHost.js
@@ -6,7 +6,11 @@ import harden from '@agoric/harden';
 import { makePrivateName } from '../util/PrivateName';
 import { allSettled } from '../util/allSettled';
 import { insist } from '../util/insist';
-import { allComparable, mustBeSameStructure } from '../util/sameStructure';
+import {
+  allComparable,
+  mustBeSameStructure,
+  sameStructure,
+} from '../util/sameStructure';
 import { makeUniAssayMaker } from './assays';
 import { makeMint } from './issuers';
 import { makeBasicMintController } from './mintController';
@@ -52,13 +56,6 @@ No invites left`;
     getInviteIssuer() {
       return inviteIssuer;
     },
-    getInviteIssuerLabel() {
-      return harden({
-        issuer: inviteIssuer,
-        description: 'contract host',
-      });
-    },
-
     // The `contractSrc` is code for a contract function parameterized
     // by `terms` and `inviteMaker`. `spawn` evaluates this code,
     // calls that function to start the contract, and returns whatever
@@ -74,17 +71,18 @@ No invites left`;
     // the installation in descriptions rather than the source code
     // itself.
     install(contractSrc) {
+      contractSrc = `${contractSrc}`;
+      const contract = evaluate(contractSrc, {
+        Nat,
+        harden,
+        console,
+        E,
+        makePromise,
+        sameStructure,
+      });
+
       const installation = harden({
         spawn(termsP) {
-          contractSrc = `${contractSrc}`;
-          const contract = evaluate(contractSrc, {
-            Nat,
-            harden,
-            console,
-            E,
-            makePromise,
-          });
-
           return E.resolve(allComparable(termsP)).then(terms => {
             const inviteMaker = harden({
               // Used by the contract to make invites for credibly
@@ -120,8 +118,18 @@ No invites left`;
               },
               redeem,
             });
-            return contract(terms, inviteMaker);
+            return contract.start(terms, inviteMaker);
           });
+        },
+        // Make it simple for the holder of an amount to verify the terms. This
+        // function will return true on success and will throw on mismatches.
+        checkAmount(allegedAmount, terms) {
+          return contract.checkAmount(allegedAmount, terms);
+        },
+        // Make it simple to verify an amount. This function will verify some of
+        // the terms, and return the unverified portion.
+        checkPartialAmount(allegedAmount, terms) {
+          return contract.checkPartialAmount(allegedAmount, terms);
         },
       });
       installationSources.init(installation, contractSrc);
--- a/core/contractHost.js
+++ b/core/contractHost.js
@@ -6,11 +6,7 @@ import harden from '@agoric/harden';
 import { makePrivateName } from '../util/PrivateName';
 import { allSettled } from '../util/allSettled';
 import { insist } from '../util/insist';
-import {
-  allComparable,
-  mustBeSameStructure,
-  sameStructure,
-} from '../util/sameStructure';
+import { allComparable, mustBeSameStructure, sameStructure } from '../util/sameStructure';
 import { makeUniAssayMaker } from './assays';
 import { makeMint } from './issuers';
 import { makeBasicMintController } from './mintController';
@@ -79,19 +75,11 @@ No invites left`;
     const installation = {};
     for (const fname of Object.getOwnPropertyNames(contractSrcs)) {
       if (typeof fname === 'string' && fname.startsWith('check')) {
-        installation[fname] = evaluateStringToFn(contractSrcs[fname]);
+        const fn = evaluateStringToFn(contractSrcs[fname]);
+        installation[fname] = (...args) => fn(installation, ...args);
       }
     }
     return installation;
-  }
-
-  function makeCheckInstallation(installation) {
-    return alleged => {
-      if (sameStructure(alleged, installation)) {
-        return installation;
-      }
-      return false;
-    };
   }
 
   /** The contract host is designed to have a long-lived credible identity. */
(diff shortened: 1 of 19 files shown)


core/contractHost.js
```diff
@@ -68,17 +71,18 @@ No invites left`;
     // the installation in descriptions rather than the source code
     // itself.
     install(contractSrc) {
+      contractSrc = `${contractSrc}`;
+      const contract = evaluate(contractSrc, {
```
katelynsills: What are the consequences of moving `evaluate` to the `install` step rather than `spawn`? We can still spawn multiple copies of the same installation with different parameters, right?

core/contractHost.js
```diff
@@ -114,9 +118,19 @@ No invites left`;
               },
               redeem,
             });
-            return contract(terms, inviteMaker);
+            return contract.start(terms, inviteMaker);
```
katelynsills: It seems like there are now three methods to call at the beginning: contractHost.install, contractHost.spawn (and spawn calls contract.start). I think adding `start` is a good choice, because contract() seems like a noun acting like a verb, but we might want extra documentation on what makes these three different.

core/contractHost.js
Chris-Hibbert: Yes.  

We needed to move contract = evaluate(...) so it would be accessible to installation.checkAmount(). Previously the only method in installation was spawn(), so it was okay that the contract was local state.

core/contractHost.js
Chris-Hibbert: Yeah, I'm working on documentation now. I think the fact that this change forced us to migrate the contracts from a pure function to a thunk is pretty unfortunate. I'm going to see if we can find a way to have the contractHost take in a function and do the extra wrapping that was required.

Requiring contract designers to write 

> const coveredCallThunk = () =>
>   const [	  harden({
>     escrowExchangeInstallationP,	    start(terms, inviteMaker) {

instead of the old 

> function coveredCall(terms, inviteMaker) {

seems like a mistake.

Chris-Hibbert:
I've done some further cleanup. It's ready for review now.

demo/contractHost/vat-fred.js
```diff
@@ -19,85 +19,36 @@ function makeFredMaker(E, host, log) {
...
-                    seatDesc: 'left',
-                  },
-                }),
+          const coveredCallTermsP = [dough10P, wonka7P, timerP, 'singularity'];
+          const verifiedSaleInvitePaymentP = E(allegedSaleInvitePaymentP)
+            .getBalance()
+            .then(balance => {
+              return E.resolve(fin55P).then(f55 =>
+                E(escrowExchangeInstallationP)
+                  .checkPartialAmount(balance, f55)
```
erights: Yeah, this one. How does fred know whether he's just verified that the left or the right is for 55 fin?

core/escrow.js
```diff
@@ -2,95 +2,144 @@
...
-    right: inviteMaker.make('right', bobSeat),
-  });
-}
-
-const escrowExchangeSrc = `(${escrowExchange})`;
-
-export { escrowExchange, escrowExchangeSrc };
+  },
+
+  checkAmount: (amount, payment) => {
```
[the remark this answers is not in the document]
erights: These parameter names are terrible. I suggest `(allegedInviteAmount, expectedTerms)`. There are many amounts running around, and invite amounts are not the first one thinks of. Here, the amount is what the caller is checking against their expectations. And this amount is obtained from an untrusted object as well.

The `expectedTerms` is how the caller expresses what they expect. Note that the caller expesses what excrowExhange installation they expect according to the receiver of the check message. However, the check method must check the alleged installation against this expectation as well, or the assurances from the rest of the checking is meaningless.

Even when checking the full terms, rather than partial checking, the caller still needs to check that the seat they'd be getting is the seat they expect. They need to provide this as an additional parameter. This would also solve the ambiguity problem with the partial checking below.

core/contractHost.js
Chris-Hibbert: The new version is back to 

> function coveredCall(terms, inviteMaker) {

But the end has changed from exporting the function to exporting a record containing labelled strings for the function sources.

> const coveredCallSrc = {
>  start: `${coveredCall.start}`,
>  checkAmount: `${coveredCall.checkAmount}`,
> };
>
> export { coveredCallSrc };

core/contractHost.js
```diff
@@ -46,79 +52,103 @@ No invites left`;
...
+    insist(typeof fn === 'function')`\n
+"${contractSrcString}" must be a string for a function, but produced ${typeof contractSrcString}`;
+    return fn;
+  }
+
+  /**
+   * Build an object containing functions with names starting with 'check' from
+   * strings in the input contract.
+   */
+  function extractCheckFunctions(contractSrcs) {
```
katelynsills: This seems a little brittle to me.

erights: Agreed. I suggested this particular hack so that we could get the rest working. Now that it is, we should figure out what we want instead. If we come up with something small and easy it can go in the PR. But it would be fine with me if we instead filed an issue on this hack now and fixed it in a subsequent PR.

core/contractHost.js
```diff
@@ -46,79 +52,103 @@ No invites left`;
...
+                seatDesc,
+              });
+              seats.init(seatIdentity, seat);
+              seatDescriptions.init(seatIdentity, seatDescription);
+              const inviteAmount = inviteAssay.make(seatDescription);
+              // This should be the only use of the invite mint, to
+              // make an invite purse whose quantity describes this
+              // seat. This invite purse makes the invite payment,
+              // and then the invite purse is dropped, in the sense
+              // that it becomes inaccessible. But it is not yet
```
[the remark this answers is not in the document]
katelynsills: This conversation was resolved, but I think this comment is still stale.

core/contractHost.js
katelynsills: Sounds good to me!

core/coveredCall.js
```diff
@@ -2,54 +2,108 @@
...
+      getRefund() {
+        return E(bobEscrowSeatP).getRefund();
+      },
+    });
+
+    return inviteMaker.make('writer', bobSeat);
+  },
+  checkAmount: (amount, terms) => {
+    const [termsLeft, termsRight, termsTimer, termsDeadline] = terms;
+    const leftAmountTerms = amount.quantity.terms[1];
```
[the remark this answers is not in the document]
Chris-Hibbert: mooted by switching coveredCall to use labelled terms

warner:
We spent some time this afternon tracking down a few of the test failures. Here are some notes:

* The code on this branch uses the deep-comparison functions in `util/sameStructure.js` to compare terms for equality. These use an `@agoric/marshal.js` internal function named `passStyleOf()` to detect promises, and arranges to compare the resolved results of those promises instead of the promises themselves.
* `passStyleOf()` was originally intended to support marshalling. It asserts that everything it sees must be frozen, because that makes marshalling stable. It also happens to be a good idea for `sameStructure.js`.
* bugs in SwingSet (https://github.com/Agoric/SwingSet/issues/95) sometimes caused the Promise returned by `E(x).foo()` to not get hardened. This will be fixed in the next release of SwingSet.
* `E(x).foo(args)` will usually harden the arguments. This used to happen immediately, but the most recent SwingSet release moved the harden to some subsequent turn after the invocation. *But* an eventual-send bug (https://github.com/Agoric/eventual-send/issues/19) meant that the harden happened immediately again. That issue has been fixed, so the next release of SwingSet should move the harden back to some subsequent turn.
* `E(x).foo(args)` will certainly harden the arguments before it tries to marshal them, which is good because marshalling will throw an exception if it gets something that isn't frozen.

We need to figure out what sorts of has-it-been-hardened behaviors we're willing to rely upon. In general, Vat code should be like Jessie code, and in Jessie code everything should be hardened before leaving the local scope. So it shouldn't be surprising that e.g. eventual-sends might harden their arguments as a side-effect, and it might be reasonable for `sameStructure.js` to do the same. And it would be polite if SwingSet/Jessie/SES-specific features, like `E()`, gave back hardened objects, so user-level code doesn't have to do that themselves. But some new features like `x!foo()` might not give us a place to perform that hardening, so we'll have to decide how aware developers need to be, and where to draw the line between "plain javascript" and our slightly-specialized SwingSet/Jessie environment.

For now, I'd say it's safest to manually `harden()` the terms objects before passing them into `sameStructure.js` functions.

warner:
BTW https://github.com/Agoric/SwingSet/issues/100 is about whether the SwingSet environment should harden things by default or not.

warner:
SwingSet-0.0.17 has been released, with the harden() changes.

Chris-Hibbert:
This seems to have sufficient consent and conversation. I'm merging it.

PR #36, opened 2019-07-17
Adds separate "use" objects and "delegated" rights

katelynsills:
## New Features
In the pixel demo, we noticed two distinct new features that were needed, that were not yet present in ERTP. 

### Use Rights as Associated Objects
First, the concept of "use" rights. Instead of a world of tokens, we wanted a world of objects that have meaningful methods (such as 'changeColor'), but which can still be handled as an ERTP payment or purse ("ERTP asset"). 

### Delegated Rights (Owner/Tenant/Tenant... Relationships)

Second, the concept of delegation. The real estate of a pixel can be leased, and the recipient of the leased pixels can also lease them to others. This can go on infinitely, in theory. We did not yet have a way to express this relationship. Moreover, since this is a common pattern with many rights, we needed a way to express this relationship in a reusable fashion.

Screen Shot 2019-08-06 at 5 53 25 PM

#### Alice can tap the faucet and change the color of the pixel:
```
const pixelPaymentP = E(gallery).tapFaucet();
const pixels = E(pixelPaymentP).getUse();
const changedAmount = await E(pixels).changeColorAll('#000000');
```
#### Alice sends Bob a delegated right (childPayment) and then revokes it

```
const pixelPaymentP = E(gallery).tapFaucet();
const childPayment = await E(pixelPaymentP).getChildPayment();
const result = await E(bob).receiveChildPayment(childPayment);
E(pixelPaymentP).revokeChildren();

```
#### Alice can get a use object from an ERTP asset, but can't get an ERTP asset from a use object
```
const pixelPaymentP = E(gallery).tapFaucet();
const pixels = E(pixelPaymentP).getUse();
```

## New Configuration Abstraction

A new configuration abstraction has been added to `makeMint` that ensures that the custom code for these two new features doesn't clutter much simpler cases like fungible mints.


core/issuers.js
```diff
@@ -123,14 +176,16 @@ Description must be truthy: ${description}`;
...
+      }
     },
     revoke(amount) {
-      this.destroy(amount);
-      return mint(amount);
+      mint.destroy(amount);
+      return mint.mint(amount);
     },
-    mint(initialBalance, _name = 'a purse') {
+    mint(initialBalance, _purseName = 'a purse') {
```
erights: Why are we still coercing `_name` on line 52 above, but no longer coercing `_purseName` here?

core/issuers.js
```diff
@@ -123,14 +176,16 @@ Description must be truthy: ${description}`;
       // for non-fungible tokens that are unique, destroy them by removing them from
       // the purses/payments that they live in
       mintController.destroy(amount);
+      if (childMint !== undefined) {
+        childMint.destroy(getChildAmount(amount)); // recursively destroy child assets
+      }
     },
     revoke(amount) {
-      this.destroy(amount);
```
erights: OMG how did a `this` sneak in? Thanks for removing it!

Separately, let's discuss better linting rules to avoid accidentally introducing `this`.

core/issuers.js
```diff
@@ -35,6 +65,21 @@ Description must be truthy: ${description}`;
...
+        const childAmount = getChildAmount(amount);
+        const childPurse = childMint.mint(childAmount);
+        const childPayment = childPurse.withdrawAll();
+        return makeUseObj(childIssuer, childPayment);
+      },
+      revokeChildren() {
+        const childAmount = getChildAmount(amount);
+        childMint.destroy(childAmount);
+      },
     });
```
erights: Just realized that the `payment` object definition is also free of side effects, so the "commit point" comment should be moved below it, here.

core/issuers.js
```diff
@@ -11,10 +11,40 @@ function makeMint(
...
+    // different issuers
+    const quantity = assay.quantity(parentAmount);
+    const childAssay = childIssuer.getAssay();
+    const childAmount = childAssay.make(quantity);
+    return childAmount;
+  }
+
   // src is a purse or payment. Return a fresh payment.  One internal
   // function used for both cases, since they are so similar.
   function takePayment(amount, isPurse, src, _name) {
```
erights: I don't see `isPurse` used below. This indicates two problems:
   * Why didn't our linting rules flag that we needed to rename the parameter `_isPurse`?
   * Doesn't our non-use mean that we might accidentally accept a payment where we were expecting only a purse, or vice versa?

core/issuers.js
```diff
@@ -35,6 +65,21 @@ Description must be truthy: ${description}`;
       getBalance() {
         return mintController.getAmount(payment);
       },
+      getUse() {
```
erights: This is the bidirectional arrow issue, whose motivation I don't yet understand.

core/issuers.js
```diff
@@ -11,10 +11,40 @@ function makeMint(
   description,
   makeMintController = makeBasicMintController,
   makeAssay = makeNatAssay,
+  makeUseObj = () => {},
```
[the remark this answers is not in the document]
erights: From the uses of `makeUseObj` below, it seems like it is always used as a two argument function. The default value, for clarity, should still be a two argument function even if both arguments are unused

```js
makeUseObj = harden((_issuer, _purseOrPayment) => harden({})),
```

core/issuers.js
```diff
@@ -81,6 +126,14 @@ Description must be truthy: ${description}`;
       const sinkPurse = issuer.makeEmptyPurse('sink purse');
       return sinkPurse.depositAll(srcPaymentP);
     },
+
+    getParentIssuer() {
```
erights: For example, `getParentIssuer` here really is the grandparent of `getChildIssuer` here, which is good.

core/issuers.js
```diff
@@ -35,6 +65,21 @@ Description must be truthy: ${description}`;
...
+        // quantity is the same, but amounts are different for
+        // different issuers
+        const childAmount = getChildAmount(amount);
+        const childPurse = childMint.mint(childAmount);
+        const childPayment = childPurse.withdrawAll();
+        return makeUseObj(childIssuer, childPayment);
+      },
+      revokeChildren() {
+        const childAmount = getChildAmount(amount);
+        childMint.destroy(childAmount);
```
erights: We have co-existing terminology `destroy`, `revoke`, and `burn`. Are they three distinct concepts? Can we consolidate some of these concepts? Can we explain their distinct meanings?

core/issuers.js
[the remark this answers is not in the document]
katelynsills: This makes sense to me. Thanks! that change is in.

core/issuers.js
```diff
@@ -11,10 +11,40 @@ function makeMint(
   description,
   makeMintController = makeBasicMintController,
   makeAssay = makeNatAssay,
+  makeUseObj = () => {},
+  parent,
```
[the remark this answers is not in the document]
katelynsills: I would agree that there are too many parameters now. I think in the case of fungible tokens, many of these new parameters are not used. We might want to created a partially applied version that has the defaults already applied. Or at least, that's how I would think of it. Is there another pattern that would work better? 

To clarify one point, `parent` is entirely optional, but it's default is undefined, so I didn't set it to undefined.

core/issuers.js
```diff
@@ -11,10 +11,40 @@ function makeMint(
...
 ) {
   insist(description)`\
 Description must be truthy: ${description}`;
 
+  const parentIssuer = parent;
```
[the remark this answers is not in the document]
katelynsills: Ah good point, I missed this

core/issuers.js
```diff
@@ -11,10 +11,40 @@ function makeMint(
...
+        makeAssay,
+        makeUseObj,
+        issuer,
+      );
+      childIssuer = childMint.getIssuer();
+    }
+    return childMint;
+  }
+
+  function getChildAmount(parentAmount) {
```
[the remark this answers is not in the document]
katelynsills: Ah good point. `parentAmount` should just be amount - the amount with the current issuer. `childAmount` is correct.

core/issuers.js
katelynsills: This purseName never gets used. I'm not sure how that happened, but it happened before this PR.

core/issuers.js
katelynsills: Sure! I think we can have an additional eslint config file that applies the Jessie linting rules and have a separate npm command that does 'npm lint-fix-jessie' or something like that

core/issuers.js
katelynsills: got it!

core/issuers.js
katelynsills: A use object has a reference to either an ERTP payment or purse (let's assume it's a payment for this example.) However, the nature of depositing a payment takes the contents of that payment, removes them, and places them in the purse. Then when we go to use the use object, it has nothing to use! Therefore, we need to be able to get a new use right from the purse that we deposited the payment into, just to maintain exactly the same rights that we had before.

katelynsills: This became apparent to me in this test, when Bob suddenly had no use rights and I couldn't figure out why. https://github.com/Agoric/ERTP/blob/d669aba76d5ca4a4e572f7f38ea2bc55b2a5add8/demo/gallery/vat-bob.js#L30

core/issuers.js
katelynsills: Good point

core/issuers.js
katelynsills: I got rid of `destroy` so now we only have `revoke` and `burn`. `Burn` is a consensual handing over of rights that are made inaccessible, and 'revoke' is nonconsensual destroying of rights.

core/issuers.js
katelynsills: The optional jessie eslint configuration is added

katelynsills:
Just adding a note that Mark noticed that there was a particular attack that could be made with this setup, in which a third-party creates a childMint by passing in a parentIssuer (issuers are widely held) but the parent is unable to revoke the children, creating something that seems revocable, but actually isn't. I haven't proven it to myself that this is an issue yet, so I'll write up a test for this case.

core/issuers.js
katelynsills: I created a new PR that adds getName() methods to actually be able to use `name` https://github.com/Agoric/ERTP/pull/41/

core/issuers.js
katelynsills: I'm not sure about the linting rules and why they didn't catch it. 
I think given that useRights and XferRights were taken out, we don't need to distinguish between purses and payments in this function anymore. https://github.com/Agoric/ERTP/pull/41/ takes isPurse out entirely

katelynsills:
>  I haven't proven it to myself that this is an issue yet, so I'll write up a test for this case.

I wrote a test for that case. I think this isn't a viable attack because there are two easy-to-use defenses. In the tests I added, Bob uses these defenses successfully against an attack by Alice.

more/pixels/gallery.js
```diff
@@ -79,227 +56,154 @@ export function makeGallery(
...
+        return useObj.changeColor(asset.getBalance(), newColor);
+      },
+      getRawPixels() {
+        const assay = issuer.getAssay();
+        const pixelList = assay.quantity(asset.getBalance());
+        return pixelList;
+      },
+      getColors() {
+        const pixelList = useObj.getRawPixels();
+        const colors = [];
```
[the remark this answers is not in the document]
katelynsills: The design challenge here is that we can't return a set because it (as far as I know) can't be serialized easily. I just pushed up a change that still returns an array, but the array contains objects with x, y, and color properties so at least the color is mapped to a pixel. Do you know if there's a better way?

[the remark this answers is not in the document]
Chris-Hibbert: I think an interface that explicitly reveals which color applies to each pixel is a fine answer.

more/pixels/gallery.js
```diff
@@ -326,15 +230,15 @@ export function makeGallery(
     });
   }
```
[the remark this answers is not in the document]
katelynsills: I think this was from when it wasn't necessarily easy or clear that ERTP would depend on SwingSet. I just pushed up a commit that changes this - let me know if this looks like what you were thinking. Actually, what do you think about making that commit a separate PR?

demo/gallery/bootstrap.js
```diff
@@ -82,56 +82,63 @@ function build(E, log) {
...
+          gallery,
+        };
+      }
+
       switch (argv[0]) {
         case 'tapFaucet': {
           log('starting tapFaucet');
-          const aliceMaker = await E(vats.alice).makeAliceMaker();
-          const gallery = makeGallery(E, log, stateChangeHandler, canvasSize);
           log('alice is made');
```
Chris-Hibbert: no longer relevant (and a few other times below.)

more/pixels/gallery.js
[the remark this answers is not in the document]
Chris-Hibbert: Yeah, that's nice. Thanks.

It would be fine as a separate PR. Doesn't really have much to do with this one. (sorry)

demo/gallery/bootstrap.js
katelynsills: Yeah, I had originally taken this out, but then I had to change the tests to match :D

core/issuers.js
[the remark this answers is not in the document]
katelynsills: I added a new issue for cleaning up `makeMint` #46. I tried to do some clean up on this branch, but I think given that there are other 'clean up' PRs that still haven't been landed, we might want to wait until those are landed before changing `makeMint`.

core/config/fungibleConfig.js
```diff
@@ -0,0 +1,23 @@
+import harden from '@agoric/harden';
+
+import { noCustomization } from './noCustomization';
+import { makeCoreMintKeeper } from './coreMintKeeper';
+import { makeNatAssay } from './assays';
+
+// Non-fungible tokens (our default for mints) do not customize
```
michaelfig: Should this be "Fungible tokens", or am I misunderstanding?

katelynsills: Ah! Good catch, you are right

core/config/coreMintKeeper.js
```diff
@@ -0,0 +1,43 @@
...
+        assets.set(asset, newAmount);
+      },
+      recordNew(asset, initialAmount) {
+        assets.init(asset, initialAmount);
+      },
+      getAmount(asset) {
+        return assets.get(asset);
+      },
+      getType() {
+        return `${type}`;
```
erights: Please don't call these types. Let's postpone the term "type" until we have more of a type system, via idl. Types will need identity. The most you could say about this method is it returns an unqualified type name, and is thereby subject to collisions. I think you want a different concept. Like purse names, these should be seen primarily as an aid to debugging rather than an inherent part of the meaning of the program.

core/config/fungibleConfig.js
```diff
@@ -0,0 +1,23 @@
+import harden from '@agoric/harden';
+
+import { noCustomization } from './noCustomization';
+import { makeCoreMintKeeper } from './coreMintKeeper';
+import { makeNatAssay } from './assays';
+
+// Fungible tokens (our default for mints) do not customize
```
erights: This text implies that we will never want to create more elaborate fungible tokens, i.e., that this will never be the base for further customizations. From the meter and fractal reserve bank, we already know that we will.

Rather, this should be described as basic fungible or core fungible or something, to fit with your overall terminology conventions for these inheritance relationships.

more/pixels/pixelMintKeeper.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+    if (paymentKeeper.has(asset)) {
+      return paymentKeeper;
+    }
+    throw new Error(
+      `asset ${asset.getName()} was not recognized as a purse or a payment`,
+    );
+  }
+
+  const pixelMintKeeper = {
```
erights: Please re-review everything for missing calls to harden. By habit, I would have written this

```js
const pixelMintKeeper = harden({
```

If not hardened here, it should at least be hardened before being returned. We won't get all of these reliably until we have a static checker for these, but we need to develop the habits of writing code that will be closer to passing such a static checker. Each one we miss is likely a security hole.

I know I have not been vigilant about this in reviews in general. We all must be.

core/config/coreMintKeeper.js
katelynsills: Hi Mark, my apologies. I know you had mentioned this exact change before, and I took it out then, but it appears to have crept back due to all of the merging (from master, from other PRs) that has happened on this branch. I'll retract my request for review and make sure that nothing like this remains.

erights: Apologies unnecessary, but thanks for the explanation.

As long as your retracting, revising, and resubmitting, could you put in necessary `harden` calls as well? Thanks.

katelynsills: Yes, of course

more/pixels/pixelMintKeeper.js
katelynsills: I've looked for the creation of object literals and added `harden` where it was lacking

katelynsills:
@erights This now has the changes to chainmail and I think it's ready for review

core/coveredCall.js
```diff
@@ -14,7 +14,7 @@ import { mustBeSameStructure, sameStructure } from '../util/sameStructure';
  * limited, and the offerer and potential acceptor are 'bob' and 'alice'
  * respectively.
  */
-const coveredCall = {
+const coveredCall = harden({
```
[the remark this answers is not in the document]
katelynsills: I think (and I could be wrong here) that we want to harden things that are exported for the module to be pure.

core/config/coreMintKeeper.js
```diff
@@ -0,0 +1,42 @@
...
+
+import { makePrivateName } from '../../util/PrivateName';
+
+export function makeCoreMintKeeper() {
+  // An asset can either be a purse or payment. An asset keeper
+  // keeps track of either all of the purses (purseKeeper) or all
+  // of the payments (paymentKeeper) and their respective amounts.
+  function makeAssetKeeper() {
+    // asset to amount
+    const assets = makePrivateName();
```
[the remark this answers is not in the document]
katelynsills: Sure, I can change that.

core/config/noCustomization.js
```diff
@@ -0,0 +1,24 @@
+import harden from '@agoric/harden';
+
+// These methods do no customization; they just return the original
+// parameter. In other configurations, these methods would be used to
```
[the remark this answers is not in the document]
katelynsills: I think it's accurate as stands - the custom methods that it speaks of are the ones added to payment/purse, etc.

more/pixels/pixelConfig.js
```diff
@@ -0,0 +1,183 @@
+import harden from '@agoric/harden';
+
+import { makePixelMintKeeper } from './pixelMintKeeper';
+import { makePixelListAssayMaker } from './pixelAssays';
+
+/**
+ * `makePixelConfigMaker` exists in order to pass in more parameters
+// than makePixelConfig allows.
```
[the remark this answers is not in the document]
katelynsills: good catch

core/coveredCall.js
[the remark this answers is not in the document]
erights: > The only way coveredCall escapes is in the exported sources. Should our style be to harden() everything, even if it won't be allowed to escape?

Just about. When in doubt, harden.

> I think (and I could be wrong here) that we want to harden things that are exported for the module to be pure.

Yes. We will want to harden enough that the Jessie static checker can verify that the module would be pure if its imports are pure. Only avoid hardening when you're confident that the Jessie static checker would be ok with it. When in doubt, harden.

There are likely to arise special cases where we need not to harden. These should align with special cases where we need to leave Jessie and program in non-Jessie SES. It would be good to comment each of those cases explicitly.

more/pixels/pixelConfig.js
```diff
@@ -0,0 +1,183 @@
...
+ * @param  {number} canvasSize=10 the size of the gallery in pixel
+ * squares across and down
+ * @param  {issuer} parentIssuer the `parentIssuer` is used when
+ * creating a revocable childPayment or childPurse, as in the
+ * landowner/tenant/subtenant pattern. In that pattern, the owner
+ * holds assets associated with the parent issuer, the tenant holds
+ * assets associated with the child issuer, the subtenant holds assets
+ * associated with the grandchild issuer, and so forth.
+ */
+function makePixelConfigMaker(makeUseObj, canvasSize = 10, parentIssuer) {
```
erights: It is a bit weird to have an optional param followed by a mandatory param. Is `parentIssuer` actually optional? If so, it should be written as `parentIssuer = undefined` or something.

more/pixels/pixelConfig.js
```diff
@@ -0,0 +1,183 @@
...
+          ...superPayment,
+          // This creates a new use object on every call. Please see
+          // the gallery for the definition of the use object that is
+          // created here by calling `makeUseObj`
+          getUse() {
+            return makeUseObj(issuer, superPayment);
+          },
+          // Mint a new payment from the child mint with the same
+          // quantity as the original payment
+          getChildPayment() {
```
erights: Given payment `p` at level 1 with balance B, does `p.getChildPayment()` mint a new child payment at level 2 with balance `B` each time `getChildPayment` is called? Does this violate an intended conservation property?

more/pixels/gallery.js
```diff
@@ -81,230 +55,168 @@ export function makeGallery(
...
+  // is used to create the "use object" that is associated with an
+  // underlying asset (purse or payment). In this case, the use object
+  // has the methods for changing the color of pixels
+
+  function makeUseObj(issuer, asset) {
+    const useObj = harden({
+      // change the color of the pixels in the amount after checking
+      // that the asset has the authority to do so.
+      changeColor(amount, newColor) {
+        insistNonEmptyAmount(issuer, amount);
```
erights: I am disturbed by the non-uniform handling of the zero case. If the amount passed in is empty, why not honor it by doing nothing?

Admittedly, this may be too hazard prone in the current design, and this non-uniform handling of the zero case may catch more bugs than it causes. But, in the upcoming linear design, assuming we distinguish zero from used up, we can avoid honoring even a zero amount on a used up asset without a special case. Until then, we should at least have a todo to revisit it at that time.

core/issuers.js
```diff
@@ -4,52 +4,69 @@
...
+
   // assetSrc is a purse or payment. Return a fresh payment.  One internal
   // function used for both cases, since they are so similar.
-  function takePayment(
-    assetSrc,
-    srcController,
-    paymentAmount,
-    unsafePaymentName,
-  ) {
+  function takePayment(assetSrc, srcKeeper, paymentAmount, unsafePaymentName) {
```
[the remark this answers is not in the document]
katelynsills: Sure, I'll change this to `name`

more/pixels/pixelConfig.js
katelynsills: Sure, that's in

more/pixels/pixelConfig.js
katelynsills: Oh gosh, thanks so much for catching this. I've changed it to check if a childPayment is created before creating one. After creation, the balances of the parent and child payment are entirely separate. If the balance of the parent changes and the user wants to create a new childPayment, they should create a new parent purse or payment and derive a new childPayment associated with it.

core/config/basicFungibleConfig.js
```diff
@@ -0,0 +1,23 @@
...
+import { makeCoreMintKeeper } from './coreMintKeeper';
+import { makeNatAssay } from './assays';
+
+// Fungible tokens (our default for mints) do not customize
+// payments, purses, etc. They use the "basic" mintKeeper (the place
+// where the amounts per purse and payment are recorded) and use the
+// "Nat" assay, in which amounts are natural numbers and use
+// substraction and addition.
+
+// This configuration and others like it are passed into `makeMint` in
```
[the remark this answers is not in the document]
katelynsills: I took out any need for passing `makeMint` `description`, so hopefully that helps here.

core/issuers.js
```diff
@@ -4,52 +4,69 @@
...
+  // Each of these methods is used below and must be defined (even in
+  // a trivial way) in any configuration
+  const {
+    makeCustomIssuer,
+    makeCustomPayment,
+    makeCustomPurse,
+    makeCustomMint,
+    makeMintKeeper,
+    makeAssay,
+  } = makeConfig(makeMint, description);
```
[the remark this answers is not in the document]
katelynsills: So the pixelConfig needs to hold internal state and isn't a pure record. What's the best way to make and pass it in as a parameter? Here, it's passed through as a function, and the object is made in makeMint. I could see an alternative, where a presence is passed in, but it seems like that would create a significant number of cross-vat calls.

core/config/coreMintKeeper.js
```diff
@@ -0,0 +1,42 @@
...
+      has(asset) {
+        return assets.has(asset);
+      },
+    });
+  }
+
+  const purseKeeper = makeAssetKeeper();
+  const paymentKeeper = makeAssetKeeper();
+
+  const mintKeeper = harden({
```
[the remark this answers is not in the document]
katelynsills: Chris wanted to rename `purseController` and `paymentController` so I changed `mintController` as well in the same ways. Can we create an issue for the naming of `mintController` to find the right name, and merge this PR in before tackling the right naming? 

I think the same thing applies to trying to optimize the packaging of the config. Can we create an issue for that and tackle it after this PR?

more/pixels/gallery.js
katelynsills: I've added a todo to look into allowing empty amounts to be colored without throwing

more/pixels/pixelConfig.js
```diff
@@ -0,0 +1,183 @@
...
+    // Lazily creates the childMint if it doesn't already
+    // exist
+    function getOrMakeChildMint(issuer) {
+      if (childMint === undefined) {
+        const makePixelConfigChild = makePixelConfigMaker(
+          makeUseObj,
+          canvasSize,
+          issuer,
+        );
+        childMint = makeMint(description, makePixelConfigChild);
```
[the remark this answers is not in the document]
erights: Based on verbal conversation, we can come back to this after this PR. Resolving.

PR #37, opened 2019-07-24
New IDL for ERTP core classes

Chris-Hibbert:
This stably represents the interface that's stable at this point. I'd appreciate reviews both on the format and content of the IDL, and of the documentation, which presents my best understanding of how ERTP is supposed to work and be used.

I stopped before spending much time on the methods for Purses and Payments, since that part seems to be undergoing significant churn.


core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+ *
+ * Currency-like assets are fungible, and don't distinguish use rights from
+ * transfer rights. The only use of a currency is to spend it. Some goods
+ * distinguish use rights from transfer rights, like a membership card; you can
+ * exercise the rights without giving them up.
+ */
+interface Purse extends (PurseOrPayment) {
+  /** Get the Issuer for this mint. */
+  getIssuer() -> (Issuer);
+  getXferBalance() -> (Amount);
```
katelynsills: Anything like XferBalance() and UseBalance() doesn't exist now. There is only getBalance()

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
-    getXferBalance() ::Amount;
-    getUseBalance() ::Amount;
-    deposit(amount ?Amount, srcPaymentP ?reveal[Promise]) ::Amount;
-    withdraw(amount ?Amount, name ?String) ::Purse;
-  }
+
+/**
+ * Payments hold verified amounts of certain rights issued by Mints. Amounts
+ * from payments can be deposited in purses, but otherwise, the entire amount is
+ * available when the payment is transferred. A payment's balance can only fall,
```
katelynsills: I think a payment's balance can't change

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+ * through the action of withdraw.
+ * 
+ * Payments can be converted to Purses by calling
+ * payment.getIssuer().makeEmptyPurse().depositAll(payment);
```
katelynsills: I like having this instruction, but getting the issuer from the payment when the payment may not be trusted is an anti-pattern, I think.

katelynsills: Maybe the payment shouldn't even have a getIssuer()

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+ */
+interface Payment extends (PurseOrPayment) {
+  getIssuer() -> (Issuer);
+  getXferBalance() -> (Amount);
```
katelynsills: only getBalance() now

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+ * PurseOrPayment is the parent type of Purse and Payment. Most interfaces only
+ * accept one of the sub-types to reduce the incidence of the common mistake of
+ * giving away a purse when you meant to hold on to the purse and give away part
+ * of the contents. deposit() and depositAll() are the two exceptions, for which
+ * the caller explicitly says to give up all contained rights, or provides an
+ * explicit amount to transfer.
+ */
+interface PurseOrPayment {
+  getIssuer() -> (Issuer);
+  getXferBalance() -> (Amount);
```
katelynsills: just 'getBalance()'

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
-  retain(remoteAmount ?reveal<RemoteIssuer.Amount>,
-         remotePaymentP ?reveal<RemoteIssuer.Payment>,
-         name ?String) ::LocalIssuer.Payment;
-  redeem(localAmount ?LocalIssuer.Amount,
-         localPayment ?LocalIssuer.Payment,
-         name ?String) ::reveal<RemoteIssuer.Payment>;
+
+/** MintController records the amounts contained by purses. */
+interface MintController {
+  /** (Non-fungible amounts only.) Ensure no purse contains amount. */
```
katelynsills: no purse or payment

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+   */
+  recordDeposit(scPayment, srcNewRightsAmount, purse, purseNewRightsAmount);
+
+  /**
+   * Record that purse is a new Purse containing srcNewRightsAmount.
+   */
+  recordMint(purse:Purse, initialAmount:Amount);
+
+  /** Return the Amount contained in purse. */
+  getAmount(purse:Purse) -> (Amount);
```
katelynsills: purse or payment

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+  /** Get the Issuer for this mint. */
+  getIssuer() -> (Issuer);
+
+  /**
+   * Create a new Purse containing the amount. Give it the specified name or "a
+   * purse".
+   */
+  mint(initialBalance:Amount, name:String = 'a purse') -> (Purse);
+
+  /** Reset all purses to contain nothing. */
```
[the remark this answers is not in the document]
Chris-Hibbert: done

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+  destroyAll();
+
+  /** (Non-fungible amounts only.) Ensure no purse contains amount. */
```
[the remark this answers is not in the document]
Chris-Hibbert: done

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+  destroy(amount:Amount);
+
+  /**
+   * Destroy any purses or payments containing amount, then issue a new Purse
+   * with that amount.
+   */
+  revoke(amount:Amount) -> (Purse);
```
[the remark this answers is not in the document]
Chris-Hibbert: I'll be happy to update when that happens.

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+}
+
+/**
+ * Purses hold verified amounts of certain rights issued by Mints. Purses can
+ * transfer part of the balance they hold in a Payment, which has a narrower
```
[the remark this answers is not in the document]
Chris-Hibbert: I mostly agree, but getBalance seems more natural than getAmount to find the contents. Although, I guess I'd expect getBalance() on a pixel purse to tell me 'three pixels' rather than identify three particular pixels, so I'm back to mostly agreeing.

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+ * interface. A purse's balance can rise and fall, through the action of deposit
+ * and withdraw, while payments don't accept deposits.
```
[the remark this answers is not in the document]
Chris-Hibbert: As we discussed, payments don't accept deposits, but their amount can be reduced by burn() or deposit(amount, payment). I'll add a little more detail here.

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+ *
+ * The primary use for Purses and Payments is for currency-like and goods-like
+ * valuables, but they can also be used to represent other kinds of rights, such
+ * as the right to participate in a particular contract.
+ *
+ * Currency-like assets are fungible, and don't distinguish use rights from
```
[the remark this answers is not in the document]
Chris-Hibbert: This will have to be updated when we get your changes in, but I want to keep the description of currency-like assets that are used by spending them to see if I can make the same point in the new design.

core/issuers.chainmail
Chris-Hibbert: agreed. This will all have to change, but there doesn't seem to be a stable target to describe yet. If this were real IDL doc, there would be comments on every method.   :-)

core/issuers.chainmail
Chris-Hibbert: as discussed, we found two ways to reduce the balance.

core/issuers.chainmail
Chris-Hibbert: The issuer isn't intended to be tightly held. 

The new purse is as trustworthy as the payment, I believe.

[part of the reason I put that line here is as a call to create a shortcut.]

core/issuers.chainmail
Chris-Hibbert: I'll write it up when it becomes stable.

core/issuers.chainmail
Chris-Hibbert: done

core/issuers.chainmail
Chris-Hibbert: Yeah, these will all have to change with your cleanup.

core/assays.chainmail
```diff
@@ -0,0 +1,104 @@
+/**
+ * Creator and validator of resource Amounts.
```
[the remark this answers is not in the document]
Chris-Hibbert: done

core/assays.chainmail
```diff
@@ -0,0 +1,104 @@
...
+ * 
+ * Amounts are the foundation for our representation of tradable
```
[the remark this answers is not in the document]
Chris-Hibbert: I like that.

core/assays.chainmail
```diff
@@ -0,0 +1,104 @@
...
+interface Assay {
+
+  /** Human readable name for the resource. */
+  getLabel() -> (String);
+
+  /** Make a new verified Amount containing the allegedQuantity. */
+  make(allegedQuantity:Quantity) -> (Amount);
+
+  /**
+   * Verify that an Amount was issued by this Assay. Notice that this
```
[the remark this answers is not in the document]
katelynsills: maybe just "created"?

core/assays.chainmail
```diff
@@ -0,0 +1,104 @@
...
+  * assay style is too awkward to use remotely. See
+  * mintTestAssay. So coerce also accepts a bare number which it
+  * will coerce to a labeled number via assay.make.
+  */
+  coerce(allegedAmount:Amount) -> (Amount);
+
+  /** Return the Quantity contained in the Amount parameter. */
+  quantity(amount:Amount) -> (Quantity);
+
+  /** Return an empty amount. Conveys no authority. */
```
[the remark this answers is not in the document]
katelynsills: I think I was thinking that brand.add(emptyAmount) means that the same emptyAmount gets returned each time, but I think you're right that === might not hold in all cases

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+   * 
+   * Labels can be used to manually construct amounts 
+   */
+  getLabel() -> (Object);
+
+  /* Get the Assay for this Issuer. */
+  getAssay() -> (Assay);
+
+  /** Make an Amount that contains the indicated quantity. */
+  makeAmount(quantity:Quantity) -> (Amount);
```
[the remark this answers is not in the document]
katelynsills: One way I've seen is making the base of the PR the branch that it depends on.

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+
+  /** Make an empty purse associated for this kind of right. */
+  makeEmptyPurse(name:String) -> (Purse);
+
+  /**
+   * Make a new Payment that has exclusive rights to the amount, only if the
+   * Purse had those rights, exclusively or not.
+   */
+  getExclusive(amount:Amount, src:Purse, name:String = 'a payment') -> (Payment);
```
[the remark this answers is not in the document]
katelynsills: I think we should enforce the distinction, even if we don't yet. I can file an issue for that

core/issuers.chainmail
katelynsills: Ah, ok, I'm a little confused because I thought we were trying to target the version that exists in your PR.

core/issuers.chainmail
katelynsills: I think we might be misunderstanding each other. The security mechanism for ensuring that I get a real payment is calling depositAll(payment) on a purse that I create from an issuer I trust. Getting the issuer from the payment doesn't provide any security, so it is an anti-pattern that we should not use. We need to get the issuer from other means for the check that depositAll() does to have meaning.

core/issuers.chainmail
katelynsills: Hmm, why do you think that this part is not stable?

core/issuers.chainmail
```diff
@@ -1,56 +1,183 @@
...
+   */
+  revoke(amount:Amount) -> (Purse);
+}
+
+/**
+ * Purses hold verified amounts of certain rights issued by Mints. Purses can
+ * transfer part of the balance they hold in a payment, which has a narrower
+ * interface. A purse's balance can rise and fall, through the action of
+ * deposit() and withdraw(), while payments don't accept deposits, but can be
+ * reduced with burn() or by passing them as a parameter to deposit().
```
erights: or `getExclusive` (aka `claim`)

core/issuers.chainmail
erights: I suggest we do target the version in Chris' PR, as it is better understood and, I expect, will get merged soon. (I still owe you comments on it.)

As Kate suggested, this PR can depend on Chris' PR, and should if it is written to agree with Chris' PR.

core/issuers.chainmail
erights: Yes. Also `getExclusive`. Unclear how to regard `destroy*` here.

core/issuers.chainmail
erights: I like this! It explains better what we're getting by separating purse from payment. A purse is what you hold onto to indicate a prior trust relationship. A payment is generally (though not always) something you're getting from someone else as an alleged payment. The APIs should be tuned towards this pattern of relationships.

I do not (yet?) conclude from this that we should remove `getIssuer` from payment. But it is the right question to ask.

core/issuers.chainmail
```diff
@@ -1,56 +1,183 @@
...
-  getLocalIssuer() ::LocalIssuer;
-  getRemoteIssuer() ::reveal<RemoteIssuer>;
-  retain(remoteAmount ?reveal<RemoteIssuer.Amount>,
-         remotePaymentP ?reveal<RemoteIssuer.Payment>,
-         name ?String) ::LocalIssuer.Payment;
-  redeem(localAmount ?LocalIssuer.Amount,
-         localPayment ?LocalIssuer.Payment,
-         name ?String) ::reveal<RemoteIssuer.Payment>;
+
+/** MintController records the amounts contained by purses. */
```
erights: MintController somehow seems more internal. Can we avoid exposing this type in issuers.chainmail ?

core/assays.chainmail
```diff
@@ -0,0 +1,91 @@
+/**
+ * Creator and validator of asset Amounts.
+ * 
+ * Amounts are the canonical description of tradable goods. They are manipulated
+ * by mints, and represent the goods and currency carried by purses and
+ * payments. They can be used to represent currency, stock, or the abstract
+ * right to participate in a particular exchange.
+ * 
+ * Assays distinguish between Quantities and Amounts. Quantities describe the
```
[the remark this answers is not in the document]
erights: I just realized that we're missing all such pass-by-copy type declarations. Rather than describing an Amount only with prose, we should have something like

```
struct Amount(Label, Quantity) {
  label :Label;
  quantity: Quantity;
}

struct Label(Description) {
  issuer :Issuer;
  description :Description;
}
```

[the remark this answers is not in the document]
erights: Do Cap'n Proto parameterized types have a way to indicate type constraints? In something like Java Generics, I would have the type parameters above extend `Comparable`.

core/issuers.chainmail
```diff
@@ -1,56 +1,183 @@
...
+ * must be compatible with the type of asset managed by the mint.
+ * 
+ * Description doesn't have to be a string, but it will only be used for its
+ * value.
+ */
+interface MintMaker {
+  makeMint(
+      description:String,
+      makeMintController:MintControllerMaker,
+	  makeAssay:AssayMaker)
```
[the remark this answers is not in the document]
erights: https://marketplace.visualstudio.com/items?itemName=xmonader.vscode-capnp

https://github.com/xmonader/sercapnp

core/contractHost.chainmail
```diff
@@ -0,0 +1,113 @@
...
+  /** Returns the actual source of the Installation. */
+  getInstallationSourceCode(installation:Installation) -> (Text);
+
+  /**
+   * Seat invitations are packaged as payments from the Invite
+   * Issuer. Redeeming an invite returns a role object with an arbitrary
+   * interface (the interface is at the discretion of the contract) which
+   * supports interaction according to the terms.
+   */
+  redeem(invite:Payment) -> (Object); // can we say more about the returned value?
```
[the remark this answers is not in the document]
katelynsills: > Use right is like membership card: while you have it you can use without consuming it. 

I think we can make the use right work however we would like - i.e. it can be consumed. All we would have to do is add that aspect to the makeUseObj function passed into makeMint.

core/assays.chainmail
```diff
@@ -0,0 +1,104 @@
+/**
+ * Creator and validator of resource Amounts.
+ * 
+ * Amounts are the foundation for our representation of tradable
+ * goods. The are the base resources manipulated by mints, and
+ * represent the goods and currency carried by purses and
+ * payments. They can be used to represent currency, stock, or the
+ * abstract right to participate in a particular exchange.
+ * 
+ * Assays distinguish between Quantities and Amounts.  An Amount is a
```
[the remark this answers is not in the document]
Chris-Hibbert: I like my latest phrasing, which says that "Quantities describe the extent of something", while "Amounts are wrappers ... that have been validated by the Assay". What do you think of this change?

core/issuers.chainmail
Chris-Hibbert: PurseOrPayment was still in flux. Now that that's gone, I've moved to the new method names and added doc.

core/issuers.chainmail
Chris-Hibbert: done.

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+  getExclusive(amount:Amount, src:Purse, name:String = 'a payment') -> (Payment);
+
+  /**
+   * Make a new Payment that has exclusive rights to all the contents of the
+   * purse.
+   */
+  getExclusiveAll(src:Purse, name:String = 'a payment') -> (Payment);
+
+  /**
+   * Burn (destroy) amount of the rights in src. If the rights are fungible,
```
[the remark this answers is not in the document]
erights: Burning is indeed as consensual as `getExclusive`, in the same way, and needs no additional text to explain that.

But your parenthetical "(destroy)" would cause confusion. It would be a fine descriptive word if we did not expect to add an API with a method `destroy` with a very distinct non-consensual meaning. We must keep the `burn` and `destroy` concepts far away from each other.

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+   * this directly removes the rights from the purse, while if the rights are
```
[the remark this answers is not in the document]
erights: The discussion here of "fungible" is wrong. There's nothing about burning that is any specific to fungible. Like `getExclusive` it applies equally to all erights. Whether the rights are fungible or not, this only removes that amount of rights from that payment, and only if that payment already holds those rights.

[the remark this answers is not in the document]
erights: It *does* remove rights from the payment. It does make those rights forever inaccessible afterwards, hence "burned". The fact that it does it by putting it into an unobservable purse which it drops is merely an implementation decision that has no place in the specification.

Once the purse is forever inaccessible, do the erights it holds still exist? If a Merkle tree falls in the forest...

core/issuers.chainmail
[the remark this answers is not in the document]
erights: Also by `getExclusive`, though I do not suggest documenting an exhaustive list. Just FYI

core/issuers.chainmail
erights: Although Kate's PR is making good progress, I continue to suggest that. Now that Chris' PR has been merged, no git hoo ha needed. Let's get this accurate for master and into master.

Kate's PR will then need to make changes to these .chainmail files, which is exactly the workflow we need. It enables us to review the changes to the spec and the implementation together.

core/issuers.chainmail
```diff
@@ -1,34 +1,174 @@
...
+interface Payment extends (PurseOrPayment) {
+  getIssuer() -> (Issuer);
+  getXferBalance() -> (Amount);
 }
 
+/**
+ * PurseOrPayment is the parent type of Purse and Payment. Most interfaces only
+ * accept one of the sub-types to reduce the incidence of the common mistake of
+ * giving away a purse when you meant to hold on to the purse and give away part
+ * of the contents. deposit() and depositAll() are the two exceptions, for which
```
erights: Exceptions how? It is the payment argument that is losing rights.

core/issuers.chainmail
erights: Since this is about to change so completely and is not exposed for use anyway, I think you can just omit it. This part of the API on master can remain undocumented until Kate's PR.

core/issuers.chainmail
Chris-Hibbert: done.

core/issuers.chainmail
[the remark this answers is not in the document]
Chris-Hibbert: okay. I dropped that part.

core/issuers.chainmail
Chris-Hibbert: I'm currently feeling that getIssuer doesn't belong in payment. I re-wrote this hint in terms of a verified issuer.

core/issuers.chainmail
Chris-Hibbert: added.

core/contractHost.chainmail
```diff
@@ -0,0 +1,164 @@
+/**
+ * ContractHost provides a way to install and run verifiable contracts.
+ *
+ * Install(source) creates Installations. Installations represent particular
+ * contract forms, which can be spawn()'d to create Contract instances
+ * initialized with specific terms. These Contracts manage invitations for
+ * multiple seats, each of which repesents a role in an interaction.
```
katelynsills: Are role and seat the same thing here?

core/contractHost.chainmail
```diff
@@ -0,0 +1,164 @@
...
+ /**
+  * The writer of the contract can provide methods to help users of the contract
+  * verify that the terms of the contract match their expectation. These methods
+  * are defined with the installation as the first parameter, so the verifiers
+  * can validate that the caller's invitation was issued by the same one. The
+  * invocation by clients should omit this parameter, as they will be supplied
+  * with a copy of the function with that information already supplied.
+  *
+  * Users usually want to validate their invitation, the terms of the deal
+  * they're attmpting to participate in, and which role/seat they are taking.
```
katelynsills: Small typo in `attempting`

core/contractHost.chainmail
erights: Yeah. I am calling the thing a "seat", and explaining the purpose of the thing as a "role". Clearer text would be more welcome.

core/issuers.chainmail
```diff
@@ -1,56 +1,152 @@
...
+  getExclusive(amount :Amount, src :Payment, name :String = 'a payment')
+    -> (Payment);
+
+  /**
+   * Make a new Payment that has exclusive rights to all the contents of src.
+   */
+  getExclusiveAll(src :Payment, name :String = 'a payment') -> (Payment);
+
+  /**
+   * Burn amount of the rights from src. If the rights are fungible, this
```
katelynsills: I think there's no fungible/non-fungible distinction here. In both cases, the rights are deposited into a purse (and therefore taken from the payment). The purse is dropped, therefore ensuring that no one has access to the rights, even the holder of the payment originally.

core/issuers.chainmail
```diff
@@ -1,56 +1,174 @@
...
+  burn(amount:Amount, src:Payment);
+
+  /** Burn (destroy) all of the rights in src. */
+  burnAll(src:Payment);
+}
+
+/**
+ * Makes Mints.
+ * 
+ * The description becomes part of the label, used by the assay to identify
```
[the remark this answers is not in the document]
katelynsills: I can't think of anything else. This looks good!

core/issuers.chainmail
```diff
@@ -1,56 +1,152 @@
...
+  /** Deposit amount from srcPayment into this purse, returning the amount. */
+  deposit(amount :Amount, srcPayment :Payment) -> (Amount);
+
+  /**
+   * Deposit all the contents of srcPayment into this purse, returning the
+   * amount.
+   */
+  depositAll(srcPayment :Payment) -> (Amount);
+
+  /** Withdrow amount from this purse into a new Payment. */
```
katelynsills: Small typo in `Withdraw`

core/assays.chainmail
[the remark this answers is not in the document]
katelynsills: Looks good to me!

core/contractHost.chainmail
Chris-Hibbert: done

core/contractHost.chainmail
katelynsills: If we use `seat` for the thing, and the purpose of the thing, I think it's a little awkward but the meaning comes through fine. The introduction of `role` really throws me, though, because it sounds like a different thing entirely whose definition I've missed.

core/issuers.chainmail
Chris-Hibbert: The fact that it's implemented by dropping a purse isn't part of the visible specification. And that approach isn't how 'burning' is done with cryptocurrencies, as I understand it.

I explain the distinction because as far as I can tell, the 'burn' metaphor only describes what happens to non-fungible values. With BitCoin, I can burn a particular UTXO, but burn doesn't apply to "1 bitcoin of value". I can figure out how to burn an invite, or a right to a particular pixel from the (blockchain) meaning of burn. It doesn't seem to me that the metaphor says what happens to fungible rights.

core/issuers.chainmail
Chris-Hibbert: done.

core/issuers.chainmail
katelynsills: In Ethereum, you burn a fungible amount by sending it to an invalid address. It's true that they don't have purses, but a purse that is dropped and accessible by no one is very close to an account that is accessible by no one. 

I found this definition of "burn" from a [slightly questionable source](https://coinsutra.com/proof-of-burn/). I think it's sound and a widely used definition: 

> Coin burn in cryptocurrencies means sending some of the coins of native cryptocurrency or some other currency to a public address from which those particular coins can never be spent because the private keys of such an address are unobtainable.

Our burn method is doing the same thing except we are using purses rather than accounts. And our burn method acts exactly the same regardless of whether the tokens are fungible or non-fungible. I think this discussion may have gotten mixed up in the discussion of `destroy` for the pixel demo, which does depend on an identifiable pixel to be destroyed and is very different than `burn` in its mechanism.

core/issuers.chainmail
Chris-Hibbert: This paragraph is gone now.

core/contractHost.chainmail
Chris-Hibbert: removed almost all uses of 'role'. I think the remaining ones look clearly like explanation and not like a technical term.

core/issuers.chainmail
Chris-Hibbert: I dropped the extraneous explanation.

Chris-Hibbert:
This has been approved.

PR #40, opened 2019-07-24
@agoric/swingset-vat: upgrade to 0.0.16

michaelfig:
This PR removes the last vestiges of the agoric-evaluate tarball.

In this case, the package-lock.json changes are desirable.

### DIFF
--- a/demo/contractHost/bootstrap.js
+++ b/demo/contractHost/bootstrap.js
@@ -95,7 +95,7 @@ function build(E, log) {
     return E(host)
       .getInstallationSourceCode(installationP)
       .then(src => {
-        log('Does source ', src, ' match? ', src === contractSrc);
+        log('Does contract source match? ', src === contractSrc);
 
         const fooInviteP = E(installationP).spawn('foo terms');
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,26 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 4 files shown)


katelynsills:
For some reason that I can't figure out, this update has introduced a split in how the logs are produced for SES and non-SES for the test case "run contractHost Demo --trivial with(out) SES"
Screen Shot 2019-07-24 at 2 51 42 PM

michaelfig:
I was looking into this, and it appears that my Babel rewriting for infix bang doesn't preserve the whitespace.  However, this in and of itself is not a problem: it's indicative of a larger inconsistency between SwingSet's SES transforms (where I introduced infix bang) and the default @agoric/evaluate behaviour.

If @agoric/evaluate is used with the same `options.transforms` as are supplied to SES, it will produce the same results.  I will add tests and work on fixing this tomorrow.

I suspect the best way would be for @agoric/evaluate to default to the same transforms we would advocate for SwingSet (probably by moving those defaults to @agoric/evaluate and have SwingSet get them from there).  People can still call `require('@agoric/evaluate').makeEvaluators(...)` if they want to opt-out of those transforms.

michaelfig:
I implemented this change for `@agoric/evaluate`, but trying to apply transforms to SwingSet's non-SES path is tricky so I left it alone.

After module rewriting is implemented, we'll need to tackle this issue head-on and fix it properly.  Until then, I'm declaring this test as overspecified (as the contract source does indeed match, regardless of how it was rewritten in advance).

PR #41, opened 2019-07-25
Clean up mintController and issuers.js

### LINKED ISSUE #39: Clean up parameter names in MintController functions
Chris' IDL PR made me realize that I hadn't named the functions and parameters in mintController very well, since I basically copied the code over when I extracted it from issuer.js. I think the parameters can be named much more clearly, and possibly the method names as well.

katelynsills:
* Renames the parameters and methods in `mintController.js` and `pixelMintController.js` for clarity
* Makes required subsequent changes to `issuers.js`
* Adds `getName()` functions to purses and payments
* removes `isPurse` parameter from `takePayment` because that was used in the context of hardcoded useRights which were taken out. 

Closes #39

### DIFF
--- a/core/issuers.js
+++ b/core/issuers.js
@@ -17,10 +17,15 @@ Description must be truthy: ${description}`;
 
   // src is a purse or payment. Return a fresh payment.  One internal
   // function used for both cases, since they are so similar.
-  function takePayment(paymentAmount, purseOrPaymentSrc, unsafePaymentName) {
+  function takePayment(
+    purseOrPaymentSrc,
+    isPurse,
+    paymentAmount,
+    unsafePaymentName,
+  ) {
     const paymentName = `${unsafePaymentName}`;
     paymentAmount = assay.coerce(paymentAmount);
-    const oldSrcAmount = mintController.getAmount(purseOrPaymentSrc);
+    const oldSrcAmount = mintController.getAmount(purseOrPaymentSrc, isPurse);
     const newSrcAmount = assay.without(oldSrcAmount, paymentAmount);
 
     const payment = harden({
@@ -28,7 +33,7 @@ Description must be truthy: ${description}`;
         return issuer;
       },
       getBalance() {
-        return mintController.getAmount(payment);
+        return mintController.getAmount(payment, false);
       },
       getName() {
         return paymentName;
@@ -39,8 +44,8 @@ Description must be truthy: ${description}`;
     // All queries above passed with no side effects.
     // During side effects below, any early exits should be made into
     // fatal turn aborts.
-    mintController.recordNewAsset(payment, paymentAmount);
-    mintController.updateAmount(purseOrPaymentSrc, newSrcAmount);
+    mintController.recordNewAsset(payment, false, paymentAmount);
+    mintController.updateAmount(purseOrPaymentSrc, isPurse, newSrcAmount);
     return payment;
   }
 
@@ -64,7 +69,7 @@ Description must be truthy: ${description}`;
     getExclusive(amount, srcPaymentP, name) {
       return Promise.resolve(srcPaymentP).then(srcPayment => {
         name = name !== undefined ? name : srcPayment.getName(); // use old name
-        return takePayment(amount, srcPayment, name);
+        return takePayment(srcPayment, false, amount, name);
       });
     },
 
@@ -72,8 +77,9 @@ Description must be truthy: ${description}`;
       return Promise.resolve(srcPaymentP).then(srcPayment => {
         name = name !== undefined ? name : srcPayment.getName(); // use old name
         return takePayment(
-          mintController.getAmount(srcPayment),
           srcPayment,
+          false,
+          mintController.getAmount(srcPayment, false),
           name,
         );
       });
@@ -99,8 +105,8 @@ Description must be truthy: ${description}`;
 
   function depositInto(purse, amount, payment) {
     amount = assay.coerce(amount);
-    const oldPurseAmount = mintController.getAmount(purse);
-    const oldPaymentAmount = mintController.getAmount(payment);
+    const oldPurseAmount = mintController.getAmount(purse, true);
+    const oldPaymentAmount = mintController.getAmount(payment, false);
     // Also checks that the union is representable
     const newPurseAmount = assay.with(oldPurseAmount, amount);
     const newPaymentAmount = assay.without(oldPaymentAmount, amount);
@@ -109,8 +115,8 @@ Description must be truthy: ${description}`;
     // All queries above passed with no side effects.
     // During side effects below, any early exits should be made into
     // fatal turn aborts.
-    mintController.updateAmount(payment, newPaymentAmount);
-    mintController.updateAmount(purse, newPurseAmount);
+    mintController.updateAmount(payment, false, newPaymentAmount);
+    mintController.updateAmount(purse, true, newPurseAmount);
 
     return amount;
   }
@@ -144,7 +150,7 @@ Description must be truthy: ${description}`;
           return issuer;
         },
         getBalance() {
-          return mintController.getAmount(purse);
+          return mintController.getAmount(purse, true);
         },
         deposit(amount, srcPaymentP) {
           return Promise.resolve(srcPaymentP).then(srcPayment => {
@@ -155,23 +161,24 @@ Description must be truthy: ${description}`;
           return Promise.resolve(srcPaymentP).then(srcPayment => {
             return depositInto(
               purse,
-              mintController.getAmount(srcPayment),
+              mintController.getAmount(srcPayment, false),
               srcPayment,
             );
           });
         },
         withdraw(amount, paymentName = 'a withdrawal payment') {
-          return takePayment(amount, purse, paymentName);
+          return takePayment(purse, true, amount, paymentName);
         },
         withdrawAll(paymentName = 'a withdrawal payment') {
           return takePayment(
-            mintController.getAmount(purse),
             purse,
+            true,
+            mintController.getAmount(purse, true),
             paymentName,
           );
         },
       });
-      mintController.recordNewAsset(purse, initialBalance);
+      mintController.recordNewAsset(purse, true, initialBalance);
       return purse;
     },
   });
--- a/core/issuers.js
+++ b/core/issuers.js
@@ -38,9 +38,6 @@ Description must be truthy: ${description}`;
       getName() {
         return paymentName;
       },
-      getType() {
-        return 'payment';
-      },
     });
 
     // ///////////////// commit point //////////////////
@@ -182,9 +179,6 @@ Description must be truthy: ${description}`;
             purseController.getAmount(purse),
             paymentName,
           );
-        },
-        getType() {
-          return 'purse';
         },
       });
       purseController.recordNew(purse, initialBalance);
--- a/core/issuers.js
+++ b/core/issuers.js
@@ -39,13 +39,8 @@ Description must be truthy: ${description}`;
     // All queries above passed with no side effects.
     // During side effects below, any early exits should be made into
     // fatal turn aborts.
-
-    mintController.recordNewPayment(
-      purseOrPaymentSrc,
-      newSrcAmount,
-      payment,
-      paymentAmount,
-    );
+    mintController.recordNewAsset(payment, paymentAmount);
+    mintController.updateAmount(purseOrPaymentSrc, newSrcAmount);
     return payment;
   }
 
@@ -102,25 +97,20 @@ Description must be truthy: ${description}`;
   const assay = makeAssay(label);
   const mintController = makeMintController(assay);
 
-  function depositInto(purse, amount, srcPayment) {
+  function depositInto(purse, amount, payment) {
     amount = assay.coerce(amount);
-    const purseOldRightsAmount = mintController.getAmount(purse);
-    const srcOldRightsAmount = mintController.getAmount(srcPayment);
+    const oldPurseAmount = mintController.getAmount(purse);
+    const oldPaymentAmount = mintController.getAmount(payment);
     // Also checks that the union is representable
-    const purseNewRightsAmount = assay.with(purseOldRightsAmount, amount);
-    const srcNewRightsAmount = assay.without(srcOldRightsAmount, amount);
+    const newPurseAmount = assay.with(oldPurseAmount, amount);
+    const newPaymentAmount = assay.without(oldPaymentAmount, amount);
 
     // ///////////////// commit point //////////////////
     // All queries above passed with no side effects.
     // During side effects below, any early exits should be made into
     // fatal turn aborts.
-
-    mintController.recordDeposit(
-      srcPayment,
-      assay.coerce(srcNewRightsAmount),
-      purse,
-      assay.coerce(purseNewRightsAmount),
-    );
+    mintController.updateAmount(payment, newPaymentAmount);
+    mintController.updateAmount(purse, newPurseAmount);
 
     return amount;
   }
@@ -181,7 +171,7 @@ Description must be truthy: ${description}`;
           );
         },
       });
-      mintController.recordNewPurse(purse, initialBalance);
+      mintController.recordNewAsset(purse, initialBalance);
       return purse;
     },
   });
--- a/core/issuers.js
+++ b/core/issuers.js
@@ -141,9 +141,9 @@ Description must be truthy: ${description}`;
       this.destroy(amount);
       return mint(amount);
     },
-    mint(initialBalance, unsafeName = 'a purse') {
+    mint(initialBalance, name = 'a purse') {
       initialBalance = assay.coerce(initialBalance);
-      const name = `${unsafeName}`;
+      name = `${name}`;
 
       const purse = harden({
         getName() {
--- a/core/issuers.js
+++ b/core/issuers.js
@@ -15,17 +15,12 @@ function makeMint(
   insist(description)`\
 Description must be truthy: ${description}`;
 
-  // src is a purse or payment. Return a fresh payment.  One internal
+  // assetSrc is a purse or payment. Return a fresh payment.  One internal
   // function used for both cases, since they are so similar.
-  function takePayment(
-    purseOrPaymentSrc,
-    isPurse,
-    paymentAmount,
-    unsafePaymentName,
-  ) {
+  function takePayment(assetSrc, isPurse, paymentAmount, unsafePaymentName) {
     const paymentName = `${unsafePaymentName}`;
     paymentAmount = assay.coerce(paymentAmount);
-    const oldSrcAmount = mintController.getAmount(purseOrPaymentSrc, isPurse);
+    const oldSrcAmount = mintController.getAmount(assetSrc, isPurse);
     const newSrcAmount = assay.without(oldSrcAmount, paymentAmount);
 
     const payment = harden({
@@ -45,7 +40,7 @@ Description must be truthy: ${description}`;
     // During side effects below, any early exits should be made into
     // fatal turn aborts.
     mintController.recordNewAsset(payment, false, paymentAmount);
-    mintController.updateAmount(purseOrPaymentSrc, isPurse, newSrcAmount);
+    mintController.updateAmount(assetSrc, isPurse, newSrcAmount);
     return payment;
   }
 
--- a/core/issuers.js
+++ b/core/issuers.js
@@ -17,10 +17,15 @@ Description must be truthy: ${description}`;
 
   // assetSrc is a purse or payment. Return a fresh payment.  One internal
   // function used for both cases, since they are so similar.
-  function takePayment(assetSrc, isPurse, paymentAmount, unsafePaymentName) {
+  function takePayment(
+    assetSrc,
+    srcController,
+    paymentAmount,
+    unsafePaymentName,
+  ) {
     const paymentName = `${unsafePaymentName}`;
     paymentAmount = assay.coerce(paymentAmount);
-    const oldSrcAmount = mintController.getAmount(assetSrc, isPurse);
+    const oldSrcAmount = srcController.getAmount(assetSrc);
     const newSrcAmount = assay.without(oldSrcAmount, paymentAmount);
 
     const payment = harden({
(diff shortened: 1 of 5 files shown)


core/issuers.js
```diff
@@ -17,16 +17,11 @@ Description must be truthy: ${description}`;
 
   // src is a purse or payment. Return a fresh payment.  One internal
   // function used for both cases, since they are so similar.
-  function takePayment(amount, isPurse, src, _name) {
```
erights: Take a look at `issuer.getExclusive`. Without the `isPurse` flag, what prevents `getExclusive` from succeeding if passed a purse?

katelynsills: Hi Mark, `isPurse` as a parameter doesn't solve this problem. I agree that `getExclusive` should error if a purse is passed in rather than a payment, but if you look at the [original code for getExclusive](https://github.com/Agoric/ERTP/blob/d577bd40a1f7061cb9cc78cc44a91ac6c31fdc51/core/issuers.js#L84), it passes in `false` for `isPurse` regardless, meaning that it will succeed even when a purse is actually passed in. I think we can find a solution to the problem, but in the meantime, I think we should take `isPurse` out.

more/pixels/pixelMintController.js
```diff
@@ -55,28 +61,38 @@ export function makeMintController(assay) {
...
+    setPixelsToAsset(paymentAmount, payment);
+    setPixelsToAsset(newSrcAmount, srcPurseOrPayment);
   }
 
-  function recordMint(purse, initialAmount) {
-    rights.init(purse, initialAmount);
-    setLocation(initialAmount, purse);
+  // a deposit (putting a payment or part of a payment into a purse)
+  // changes the amounts of both the purse and the payment
+  function recordDeposit(payment, newPaymentAmount, purse, newPurseAmount) {
```
Chris-Hibbert: I can imagine that we'd do some extra record keeping or balance checking to take advantage of the distinctions between recordNewPurse, recordNewPayment, and recordDeposit, but we're not doing any of that. It seems like the interface would be simpler if we just had two functions: 
updateBalance(purseOrPayment, newBalance)
and 
recordNewAsset(purseOrPayment, initialBalance).

Otherwise, it doesn't seem like the distinctions between the three current functions is gaining us much.

more/pixels/gallery.js
```diff
@@ -245,9 +245,13 @@ export function makeGallery(
...
 
-    pixelMint.destroy(pixelAmount);
-    useRightMint.destroy(useRightAmount);
-    transferRightMint.destroy(transferRightAmount);
+    try {
+      pixelMint.destroy(pixelAmount);
+      useRightMint.destroy(useRightAmount);
+      transferRightMint.destroy(transferRightAmount);
+    } catch (err) {
+      console.log(err);
```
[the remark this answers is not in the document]
katelynsills: Before this PR, when destroy happened and the amount wasn't found, it just returned with no error. I changed that logic to throw an error, but that necessitated a change here to handle it gracefully (this happens every time the gallery tries to revoke a pixel at the start). 

All of these lines go away in https://github.com/Agoric/ERTP/pull/36 so this should be a very temporary change to handle the exceptions that won't exist in the next few days.

more/pixels/pixelMintController.js
```diff
@@ -12,40 +13,45 @@ export function makeMintController(assay) {
...
+    const purseOrPayment = pixelToPursePayment.get(strPixel);
     // amount is guaranteed to be there
     // eslint-disable-next-line no-use-before-define
     amount = assay.coerce(amount);
-    const srcOldRightsAmount = rights.get(location);
-    // eslint-disable-next-line no-use-before-define
-    const srcNewRightsAmount = assay.without(srcOldRightsAmount, amount);
+    const originalAmount = rights.get(purseOrPayment);
+
+    // esline-disable-next-line no-use-before-define
```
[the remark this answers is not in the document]
katelynsills: Assay is not yet defined, but I've removed this rule for the whole file, so we can take out this line if we want to

core/mintController.js
```diff
@@ -20,30 +15,44 @@ export function makeBasicMintController() {
...
+  function recordNewPayment(
+    srcPurseOrPayment,
+    payment,
+    paymentAmount,
+    newSrcAmount,
   ) {
-    rights.set(srcPayment, srcNewRightsAmount);
-    rights.set(purse, purseNewRightsAmount);
+    rights.init(payment, paymentAmount);
+    rights.set(srcPurseOrPayment, newSrcAmount);
```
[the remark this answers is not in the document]
katelynsills: Good point!

more/pixels/gallery.js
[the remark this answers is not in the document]
katelynsills: So I think in answering the questions, I would have to answer in the context of both this change, and the change to `pixelMintController.destroy`:

try-catch with a log message will log that the amount didn't exist, where previously, there was no log message and no exception.

It doesn't prevent execution from stopping, and neither did the code before this PR

It makes the exceptions more visible in debugging (because they didn't exist before), and for normal running and our test framework, it is the same (because normal running isn't affected by the logs, and our test framework uses a different logging tool than console.log for the test logs)

more/pixels/pixelMintController.js
```diff
@@ -12,40 +13,45 @@ export function makeMintController(assay) {
   // pixel to purse/payment
   const pixelToPursePayment = new Map();
 
-  function setLocation(amount, pursePayment) {
-    // purse/payment is the key of rights
+  function setPixelsToAsset(amount, purseOrPayment) {
```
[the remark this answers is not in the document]
katelynsills: I like purseOrPayment, but it's so long that I was afraid of getting called out for long names :). What do you think of Asset?

more/pixels/pixelMintController.js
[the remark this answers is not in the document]
katelynsills: Whoops, I thought this was in issuer.js for some reason. You're right!

more/pixels/pixelMintController.js
katelynsills: Hmm,  creating a new Payment and creating a new Purse does different things, though. Were you imagining that this logic would get pushed to issuer.js and mintController would be very simple?

katelynsills: I'm trying this out, and I think I like it

core/issuers.js
erights: Looked. Called from `getExclusive` it is always false, so only payments are accepted. Called from `withdraw`, it is always true, so only purses are accepted.

katelynsills: Mark, can you explain how this prevents a payment from actually being passed in in place of a purse and vice versa?

erights: In the original you link to, at https://github.com/Agoric/ERTP/blob/d577bd40a1f7061cb9cc78cc44a91ac6c31fdc51/core/issuers.js#L36 is the code

```js
    if (isPurse) {
      insist(useRights.has(src))`\
Purse expected: ${src}`;
    } else {
      insist(homePurses.has(src))`\
Payment expected: ${src}`;
    }
```

In that original, only purses were registered in `useRights`. Only payments were registered in `homePurses`. Of course, when we lost `useRights` and `homePurse` we lost that too, so we need another way to distinguish these. The remaining PrivateName, `rights`, has been abstracted into MintController, whose API names also claim to distinguish between purses and payments, but use one `rights` table for both.

I think the cleanest way to fix this is for MintController to have two PrivateNames, `purseRights` and `paymentRights`, so that only purses are found in the first and only payments are found in the second. Avoid using one brand to certify two different things.

katelynsills: Sure, sounds good. I'll add that

more/pixels/gallery.js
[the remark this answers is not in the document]
Chris-Hibbert: thanks. That seems fine, then.

more/pixels/pixelMintController.js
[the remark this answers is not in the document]
Chris-Hibbert: I don't mind 'Asset', as long as we're consistent. But now with the decision to enforce the distinction between Purses and Payments, perhaps we won't need a new word for both?

core/issuers.js
katelynsills: That's added. @Chris-Hibbert  I had to make some changes to the corkboard tests [here](https://github.com/Agoric/ERTP/blob/4776c4c3a555e05d236f5608d627f918f0198402/demo/gallery/vat-bob.js#L72) because sometimes purses were being passed as payments and vice versa.

more/pixels/pixelMintController.js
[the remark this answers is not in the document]
katelynsills: There are still arguments that are either a purse or payment, for instance in `takePayment`, so I think we still have this problem even with `rights` being split. I just pushed up a commit that uses 'asset' instead of 'purseOrPayment'. Let me know what you think!

katelynsills:
@Chris-Hibbert I think I've taken out 'isPurse' entirely in a way that still maintains the security of purses versus payment. Happy to talk about whether it still smells :) or if there's a better way to go about it

more/pixels/pixelMintController.js
```diff
@@ -44,66 +63,24 @@ export function makeMintController(assay) {
...
-    // we don't know if this is a purse or payment, so handle both cases
-    if (purses.has(asset)) {
-      const originalAmount = purses.get(asset);
-      const newAmount = assay.without(originalAmount, amount);
-      purses.set(asset, newAmount);
-      // Reset the mappings from everything in the amount to the purse
-      // or payment that holds them.
-      recordPixelsAsAsset(newAmount, asset);
-    }
+    const type = asset.getType();
```
Chris-Hibbert: This seems to be the only place where the code doesn't already know the type of the asset. It's the only place that asks for the type (which is another smell). 

Since the controllers are internal implementation details of the pmController, I think it would be clean enough for it to ask both controllers for the amount, and call updateAmount on whichever responded. I'm sure there are other ways to do it; if you find another approach that you like better (without asking the asset for its type), that would be fine, too.

core/mintController.js
```diff
@@ -4,50 +4,36 @@ export function makeBasicMintController() {
...
+      updateAmount(asset, newAmount) {
+        assets.set(asset, newAmount);
+      },
+      recordNew(asset, initialAmount) {
+        assets.init(asset, initialAmount);
+      },
+      getAmount(asset) {
+        return assets.get(asset);
+      },
+      destroyAll() {
```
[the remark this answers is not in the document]
katelynsills: Good idea. It's outside because it acts on both purses and payments. I'll make a comment

core/issuers.js
```diff
@@ -17,30 +17,38 @@ Description must be truthy: ${description}`;
...
+  function takePayment(
+    assetSrc,
+    srcController,
+    paymentAmount,
+    unsafePaymentName,
+  ) {
     const paymentName = `${unsafePaymentName}`;
     paymentAmount = assay.coerce(paymentAmount);
-    const oldSrcAmount = mintController.getAmount(assetSrc, isPurse);
+    const oldSrcAmount = srcController.getAmount(assetSrc);
```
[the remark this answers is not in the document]
katelynsills: You're right that we don't need it in this call, but we do need it down below in the same function, and using srcController here will fail fast if it is the wrong asset type, so I'm kind of leaning towards this. What do you think?

more/pixels/pixelMintController.js
katelynsills: I removed the `getType` from purses and payments and added a `has` method to assetController instead that allows us easily define predicates `isPurse()` and `isPayment()` that can be used in the case of `destroy(amount)`. Does this solve the problem better?

core/issuers.js
```diff
@@ -17,30 +17,38 @@ Description must be truthy: ${description}`;
...
         return issuer;
       },
       getBalance() {
-        return mintController.getAmount(payment, false);
+        return paymentController.getAmount(payment);
       },
       getName() {
         return paymentName;
       },
+      getType() {
```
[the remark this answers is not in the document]
katelynsills: Good point, I took this out entirely

more/pixels/pixelMintController.js
Chris-Hibbert: As we discussed, this is a big improvement.

If we're going to expose something to help users find out the type of an asset, I prefer the getType() that returns a string rather than a boolean, since that's better for people and harder for code to use.

katelynsills: Ok, there is now a getController(asset) method for pixelMintController

core/issuers.js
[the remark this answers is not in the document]
erights: AFAICT, the fail-fast here is required. The key to assaying the alleged src parameter is that it is found as a key in the appropriate encapsulated table, where only genuine srcs of that kind are found.

If you did instead

```js
const oldSrcAmount = assetSrc.getBalance();
```

then Alice could use a misbehaving payment to seem to pay Bob without actually paying him.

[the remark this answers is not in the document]
erights: Btw, I think like the approach you took, or using different controller types rather than the `isPurse` flag. However, I would like for both controllers to be internal mechanism only, not exposed as API that needs documenting. If they are exposed, making two of them increases the API surface yet more.

These are points to discuss later, after this PR is merged.

core/issuers.js
```diff
@@ -53,21 +62,23 @@ Description must be truthy: ${description}`;
...
     },
 
-    getExclusiveAll(srcPaymentP, name = 'a payment') {
-      return Promise.resolve(srcPaymentP).then(srcPayment =>
-        takePayment(
-          mintController.getAmount(srcPayment),
-          false,
+    getExclusiveAll(srcPaymentP, name) {
+      return Promise.resolve(srcPaymentP).then(srcPayment => {
+        name = name !== undefined ? name : srcPayment.getName(); // use old name
```
erights: Btw, this is a great example of where the proposed "null coalescing syntax" that just advanced would come in handy. You could correctly rewrite this to

```js
name = name ?? srcPayment.name();
```

which differs only in that it will treat `null` as it treats `undefined`, which is harmless. The way most people (likely including me when I'm not paying enough attention) would code this today is

```js
name = name || srcPayment.getName();
```

which would prevent purposely choosing the empty string as a name. The problem is that the empty string is falsy, which is what `||` tests.

core/issuers.js
```diff
@@ -116,28 +124,32 @@ Description must be truthy: ${description}`;
...
       // the purses/payments that they live in
-      mintController.destroy(amount);
+      destroy(amount);
     },
     revoke(amount) {
       this.destroy(amount);
       return mint(amount);
     },
-    mint(initialBalance, _name = 'a purse') {
+    mint(initialBalance, unsafeName = 'a purse') {
```
erights: We're normally suspicious of incoming parameters. Sometimes we make this explicit in the parameter names, like `allegedFoo`, to really draw our vigilance to examining how we use it. For `unsafeName`, you coerce it to a string immediately. Afterwards there's nothing more to worry about. I'd just name the parameter `name` and turn the `const name` declaration below into an assignment.

core/mintController.js
```diff
@@ -1,50 +1,57 @@
...
-    srcPayment,
-    srcNewRightsAmount,
-    purse,
-    purseNewRightsAmount,
-  ) {
-    rights.set(srcPayment, srcNewRightsAmount);
-    rights.set(purse, purseNewRightsAmount);
-  }
+  const purseController = makeAssetController('purse');
+  const paymentController = makeAssetController('payment');
```
erights: I really like that there's no code difference between a purseController and a paymentController.

core/mintController.js
```diff
@@ -1,50 +1,57 @@
...
+    return purseController.has(asset);
   }
 
-  function getAmount(pursePayment) {
-    return rights.get(pursePayment);
+  function isPayment(asset) {
+    return paymentController.has(asset);
   }
 
   const mintController = {
```
erights: Curious why you're defining the methods by defining functions first and then gathering them into properties. If they call each other, that could be a good reason. Otherwise, my inclination is to define them inside the object using concise method syntax. This is (slightly) semantically better too, in that methods (functions defined using concise method syntax) have no [[Construct]] behavior, i.e., you can't `new` them.

more/pixels/pixelMintController.js
```diff
@@ -1,95 +1,114 @@
...
+    insist(pixelList.length === 1)`amount must contain exactly one pixel`;
 
     const pixel = pixelList[0];
     const strPixel = getString(pixel);
-    if (!pixelToPursePayment.has(strPixel)) {
-      return;
-    }
-    const location = pixelToPursePayment.get(strPixel);
+    insist(pixelToAsset.has(strPixel))`\
+      pixel ${strPixel} could not be found to be destroyed`;
```
erights: My practice is to not have any leading whitespace on the line after backquote-backslash. We should discuss which makes code more or less readable. I see both sides. But we should try to consistently use one style until we change our mind to consistently use another style.

Also, the indentation here is a semantic change. These leading spaces are part of the literal text passed along to `insist`.

core/issuers.js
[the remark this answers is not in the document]
katelynsills: Hi Mark, what do you mean by internal mechanisms vs exposed as API? Does something being in a separate module file make it exposed as an API?

core/issuers.js
katelynsills: Nice! I'm looking forward to being able to use that syntax. Just to make sure, is it ok as is until then?

core/mintController.js
katelynsills: Hmm, I think this might be a misunderstanding. All of the methods are defined using concise method syntax. `isPurse()` and `isPayment()` here aren't object methods so they're just defined as functions.

more/pixels/pixelMintController.js
katelynsills: Good point. I'll make that change

core/mintController.js
katelynsills: Awesome!

core/issuers.js
erights: Yes.

core/issuers.js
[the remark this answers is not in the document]
erights: Can one create a useful new mint without having to think about mintControllers. I see now that my concern is likely already satisfied enough. Double checking:
   * For simple uses, the caller of makeMint would just have this argument position default, in which case `makeBasicMintController` would be chosen.
   * `makeBasicMintController` is fine for all normal uses that do not need a pre-emption hierarchy.
   * Other than this parameter position, mintControllers are not externally visible, or even mentioned, anywhere the API surface we do expose.

I am not concerned about the separate file. But one thing we should keep in mind as we do safe module and package support for SES is that a package should be able to encapsulate internal modules. I don't know that we would choose to encapsulate `./mintController`, but it will be good to have that choice!

core/mintController.js
```diff
@@ -1,50 +1,57 @@
...
-  // purse/payment to amount
-  let rights = makePrivateName();
-
-  function recordPayment(src, payment, amount, srcNewRightsAmount) {
-    rights.set(src, srcNewRightsAmount);
-    rights.init(payment, amount);
+  // An asset can either be a purse or payment. An asset controller
+  // keeps track of either all of the purses (purseController) or all
+  // of the payments (paymentController).
+  function makeAssetController(type) {
```
erights: Don't call this `type`. At some point this code will grow to have types in some programming sense of the word type, so we should avoid it for other uses.

core/mintController.js
```diff
@@ -1,50 +1,57 @@
...
+      updateAmount(asset, newAmount) {
+        assets.set(asset, newAmount);
+      },
+      recordNew(asset, initialAmount) {
+        assets.init(asset, initialAmount);
+      },
+      getAmount(asset) {
+        return assets.get(asset);
+      },
+      getType() {
```
erights: Likewise, don't call it `getType`

core/mintController.js
erights: I see. My criticism applied to the prior version but I had not appreciated that it is mostly fixed in this one. However, why is `destroy` a function rather than a concise method?

core/issuers.js
Chris-Hibbert: I'm usually skeptical of re-using variable names like this. Is there a reason it's safer in javascript than in other languages?

I think the usual guideline is that it's bad practice to change the value assigned to a variable so readers who skim the code might miss the change. I strongly prefer using a name like unsafeFoo, and declaring a new variable to hold the cleaned-up value.

more/pixels/pixelMintController.js
```diff
@@ -1,95 +1,115 @@
...
-  }
-
-  function getAmount(pursePayment) {
-    return rights.get(pursePayment);
+    // delete pixel from pixelToAsset
+    pixelToAsset.delete(pixel);
   }
 
   const mintController = {
     destroy,
```
erights: Same question about whether this `destroy` should be a method or a function.

core/issuers.js
katelynsills: I agree with Chris in that I'd prefer not to re-assign to a variable.

core/mintController.js
katelynsills: Destroy takes in an amount and acts on both purses and payments (potentially in the same call) depending on where the pixels in the amount was located, so there wasn't an easy way to contain it within a purseController or paymentController.

core/issuers.js
erights: What about line 51 below?

Almost always in defensive programming, all arguments are things we're suspicious of, whether we call this unsafe* or alleged* notReliedOn* or something else. I am certainly in favor of any convention that helps us remember to be suspicious of these! However, it needs to be a convention we can apply consistently. 

With `alleged*` I chose only things for which we must remain suspicious event after the beginning of the function tries to normalize and validate. This, initialBalance on line 141 is not `allegedInitialBalance` or `unsafeInitialBalance`.

By what principle is `name` worth prefix but `initialBalance` not?

Regarding declaring new vs reassigning to old, I'm calm about it. We need a convention. But fixing line 142 and not 141 is jarring. ***If*** we don't come up with a naming convention for labeling all parameters we can't rely on, and ***if*** we continue to use unqualified names for most of these, then I would prefer reassignment for normal-named parameter variables that are immediately normalized to relied-upon values on entry to the function, before it does anything else. Otherwise, I agree that declaring new variables is likely better.

We should find principles to create conventions we can widely adhere to.

core/mintController.js
katelynsills: Ah, thanks for explaining. I could definitely make this a concise method. Would you like for that change to be part of this PR?

erights: y

core/issuers.js
katelynsills: Sounds good to me. I can create an issue for this to push the discussion out of this PR.

PR #43, opened 2019-07-25
Rename and move the files in 'demo'

### LINKED ISSUE #42: Move "demo" folder to be under tests
We have some vestigial organization leftover from when ERTP was part of SwingSet. Let's rename the 'demo' folder and move it under tests, since it is only used when testing.

katelynsills:
This PR divides the tests into unit tests and tests that involve Swingset. It also moves the scaffolding needed for the tests (the vat files and so forth) under a folder called `scaffolding` under `SwingsetTests/`. 

It eliminates the `pixel-demo` tests that were written before we had the concept of a gallery, since those tests are now subsumed by the gallery tests. 

I tried to organize this into meaningful commits, but let me know if you need me to break it up further. 

Closes #42


test/swingsetTests/test-contractHost.js
```diff
@@ -0,0 +1,179 @@
+import { test } from 'tape-promise/tape';
```
Chris-Hibbert: I think these test-*.js files should be in the same directory as the vats they rely on. There doesn't seem to be a good reason to separate them from the code they rely on.

So this file would be test/swingsetTests/scaffolding/contractHost/ along with bootstrap and vat-alice, vat-bob, and vat-fred.

test/swingsetTests/scaffolding/contractHost/vat-host.js
```diff
@@ -3,7 +3,7 @@
 import harden from '@agoric/harden';
 import evaluate from '@agoric/evaluate';
 
-import { makeContractHost } from '../../core/contractHost';
+import { makeContractHost } from '../../../../core/contractHost';
 
 function setup(syscall, state, helpers) {
```
Chris-Hibbert: I wonder whether there's a simple way to have a master copy of vat-host and vat-mint, and have all the tests that need them share it rather than having separate copies in every directory.

katelynsills: What do you think about adding this as an issue, but not doing it in this PR?

test/swingsetTests/test-contractHost.js
katelynsills: Sounds good!

katelynsills:
> I think .../scaffolding/... is an unneeded level in the hierarchy. Each of the swingsetTests can be a single directory with the test-*.js file and all the vat code.

Good idea. Scaffolding is no more

PR #58, opened 2019-08-01
A test case for presence corruption with the eventual send candidate.

Chris-Hibbert:
This fails with the new eventual send library, and succeeds in the current state.


package-lock.json
```diff
@@ -28,6 +28,13 @@
         "@agoric/transform-bang": "^0.3.1",
         "@babel/generator": "^7.5.5",
         "esm": "^3.2.5"
+      },
```
katelynsills: what do the package-lock.json changes do?

Chris-Hibbert: I thought I had reverted this. Let me try again.
The bug this test case is meant to reveal is provoked by the current draft of the eventual-send code. 

The test case is intended to succeed in the current master, but be ready to fail if the bug in the draft eventual-send code isn't fixed.

test/presenceCorruption/test-demos.js
```diff
@@ -0,0 +1,27 @@
+import { test } from 'tape-promise/tape';
```
katelynsills: Can we rename this from test-demos.js?

Chris-Hibbert: Oh yeah. done.

Chris-Hibbert:
approved by all; merging

PR #60, opened 2019-08-01
adds commands to run jessie eslint rules

katelynsills:


katelynsills:
Ok, this should not have any `eventual-send` changes now @Chris-Hibbert

PR #72, opened 2019-08-08
Repair chainmail Files

Chris-Hibbert:
The IDL doesn't allow structs to declare that they extend another. I'll have to see if it makes sense to add that. For now, we'll comment out the declarations.


Chris-Hibbert:
Explicit approval from michaelfig; implicit approval from erights.

PR #76, opened 2019-08-09
Enforce payment linearity

### LINKED ISSUE #63: Payment Linearity
General principle: Payments are either new or dead

1. Add two methods on `issuer`: `combine` (two payments go in, they become dead and a new one goes out) and `split` (one payment goes in, it becomes dead, and two new ones are created)
2. Remove `deposit` - only use `depositAll`
3. Remove `getExclusive` -> only use `getExclusiveAll`
4. Remove `burn` - only use `burnAll`

katelynsills:
Closes #63 

This PR enforces "payment linearity". 

General principle: Payments are either new or dead

1. Add two methods on `issuer`: `combine` (two payments go in, they become dead and a new one goes out) and `split` (one payment goes in, it becomes dead, and two new ones are created)
2. Remove `deposit` - only use `depositAll`
3. Remove `getExclusive` -> only use `getExclusiveAll`
4. Remove `burn` - only use `burnAll`


core/issuers.chainmail
```diff
@@ -21,21 +21,36 @@ interface Issuer (Amount (Quantity)) {
...
+  combine(paymentsArray :List(Payment), name :String = 'combined
+  payment') -> (Payment)
+
+  /** 
+   * Split a single payment into multiple payments, according to the
+   * amounts and names passed in. 
+   */
+  split(payment, amountsArray :List(Amount), namesArray
+  :List(String)
+    -> (List(Payment))
```
Chris-Hibbert: The return type should be outside the parens for the parameters.

`
 split(payment, amountsArray :List(Amount), namesArray  :List(String))
    -> (List(Payment)
`

core/issuers.js
```diff
@@ -15,6 +15,35 @@ function makeMint(
   insist(description)`\
 Description must be truthy: ${description}`;
 
+  // The payment object is created in multiple functions (takePayment,
+  // split, combine), so we extract its particulars here and call this
```
Chris-Hibbert: Would you mind adding that the payment's amount is recorded in controller/keeper, and not known to the payment?

core/issuers.js
```diff
@@ -66,10 +85,67 @@ Description must be truthy: ${description}`;
...
+      paymentController.recordNew(combinedPayment, totalAmount);
+      return combinedPayment;
+    },
+
+    split(payment, amountsArray, namesArray) {
+      namesArray =
+        namesArray !== undefined
+          ? namesArray
+          : Array(amountsArray.length).fill('a split payment');
+
```
Chris-Hibbert: We need a check that the namesArray (at this point) is the same length as the amountsArray.

test/unitTests/core/test-issuers.js
```diff
@@ -0,0 +1,102 @@
...
+      payments.push(purse.withdraw(1));
+    }
+
+    const combinedPayment = issuer.combine(payments);
+    t.deepEqual(combinedPayment.getBalance(), issuer.makeAmount(100));
+    for (const payment of payments) {
+      t.throws(() => payment.getBalance());
+    }
+
+    const badAmountsArray = Array(2).fill(issuer.makeAmount(10));
```
Chris-Hibbert: Please start two new tests here. (goodsplit and badsplit) 
 Each test case should test a single failure mode when possible. The split tests don't depend on the input payment being one that was just built with combine.

test/unitTests/core/test-issuers.js
```diff
@@ -0,0 +1,102 @@
...
+    const purse = mint.mint(1000);
+    const targetPurse = issuer.makeEmptyPurse();
+    const payment = await purse.withdraw(7);
+    try {
+      await targetPurse.depositExactly(issuer.makeAmount(6), payment);
+      t.fail();
+    } catch (err) {
+      t.equal(err.message, 'payment balance must equal amount');
+    }
+    await targetPurse.depositExactly(7, payment);
```
Chris-Hibbert: similarly, break this up here. good deposit and bad deposit can be separate test cases.

[and below.]

core/issuers.chainmail
katelynsills: Good catch

core/issuers.js
katelynsills: Yup, that sounds good

test/unitTests/core/test-issuers.js
katelynsills: Good idea

core/issuers.js
katelynsills: added

katelynsills:
Do not merge until 0.0.11

core/issuers.chainmail
```diff
@@ -21,21 +21,36 @@ interface Issuer (Amount (Quantity)) {
   /** Make an empty purse associated with this kind of right. */
   makeEmptyPurse(name :String) -> (Purse);
 
+  /** Combine multiple payments into one payment. */
+  combine(paymentsArray :List(Payment), name :String = 'combined
```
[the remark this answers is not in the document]
katelynsills: good catch. that should be fixed, let me know if it needs more

katelynsills:
I realized that we need to change `depositExactly/All`, `claimExactly/All` and `burnExactly/All` to also kill the payment involved for full payment linearity. I pushed up some new commits that haven't been reviewed yet.

PR #77, opened 2019-08-09
WIP: Scooter, atomic trade of remote assets.

erights:
No one should feel obligated to look at this yet. But if you're curious....

This differs some from what we sketched on the whiteboard, but preserves the essence.

Our Chainmail idl really helped, simply as a way to document the public interface, and keep that documentation separate from implementation comments. If you do look, I recommend looking at the **`scooter.chainmail`** first.

The sentry is intended to provide exit conditions that can be validated. It doesn't yet, but I think it is close.

I have not yet even run it, much less tested anything. The scooter.chainmail file has never been parsed. The real test is how awkward the API is to use. I suspect that the current answer is "too awkward". How much of this is reform of scooter itself, and how much is support libraries for using scooter, remains to be seen. I suspect it will be some of both.

My hope is that this is a good enough proof of concept that, like the early ERTP, we can proceed from here by incremental improvements.


erights:
Obsolete

PR #83, opened 2019-08-12
WIP: Pegs preserve original label

erights:
WIP: Not yet a candidate, as it has not yet been run or tested. However, I'm posting this now as 
   * This updates `makePeg` to be compatible with our new `makeConfig`-based system for making mints and issuers.
   * Scooter is about to depend on this change in behavior.

The peg's new local assay uses the label of the remote issuer, so the amounts it makes are labeled as from that issuer. They are not endorsed by that issuer, but they wouldn't be anyway since they are distant and pass-by-copy. Rather, they coerce to amounts accepted by the remote issuer.

This has pros and cons. 
   * The pros help use of a local peg merely as a bookkeeping advice to track updates to a remotely held escrow purse, as is done in scooter. 
   * The cons are that abstractions, like the escrow exchange that are parameterized by amounts, rather than (issuer,amount) pairs, will extract the intended issuer from the amount, routing around the peg's issuer. Sometimes this will even be intended. When not intended, we'll need to pass (issuer,amount) pairs instead.

### DIFF
--- a/core/issuers.js
+++ b/core/issuers.js
@@ -125,7 +125,7 @@ Description must be truthy: ${description}`;
   let label = harden({ issuer, description });
   const assay = makeAssay(label);
   label = assay.getLabel();
-  
+
   const mintKeeper = makeMintKeeper(assay);
   const { purseKeeper, paymentKeeper } = mintKeeper;
 
--- a/core/issuers.js
+++ b/core/issuers.js
@@ -120,9 +120,12 @@ Description must be truthy: ${description}`;
   // additional methods
   const issuer = makeCustomIssuer(coreIssuer);
 
-  const label = harden({ issuer, description });
-
+  // This module only suggests a label to makeAssay. But the actual
+  // label is what makeAssay says it is.
+  let label = harden({ issuer, description });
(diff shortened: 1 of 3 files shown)


Chris-Hibbert (changes requested):

core/issuers.js
```diff
@@ -120,9 +120,12 @@ Description must be truthy: ${description}`;
   // additional methods
   const issuer = makeCustomIssuer(coreIssuer);
 
-  const label = harden({ issuer, description });
-
+  // This module only suggests a label to makeAssay. But the actual
+  // label is what makeAssay says it is.
+  let label = harden({ issuer, description });
```
katelynsills: Maybe if we called this "suggestedLabel" and then called what comes out of makeAssay "label" we might be able to get rid of the comment.

katelynsills:
> The peg's new local assay uses the label of the remote issuer, so the amounts it makes are labeled as from that issuer.

I think I don't understand the benefit we're getting. It seems confusing to fake a label rather than just use the local label and map it to the remote.

erights:
Obsolete

PR #88, opened 2019-08-14
update readmes with better and update-to-date explanations

katelynsills:


GALLERY-README.md
```diff
@@ -6,43 +6,122 @@ number of additional features that showcase the unique affordances of
...
+to someone else. 
+
+## Creating Pixels
+
+ERTP provides a `makeMint` function that takes in a configuration
+function. The pixel demo configuration implements the hierarchical
+rights and allows us to add additional methods to purses and payments.
+
+The configuration function passed to `makeMint` has a slot for an
+assayMaker function. An assay describes how amounts of a particular
```
erights: `assayMaker`

Check for other missing backquotes.

README.md
```diff
@@ -1,13 +1,120 @@
...
+[issuers.js](core/issuers.js) takes
+in an optional configuration that allows for many more possibilities. 
+
+## Pixel Gallery Demo
+
+To explore further types of digital assets, beyond fungible tokens, we
+decided to create a demo for our
+[testnet](https://github.com/Agoric/cosmic-swingset) in which users
+can buy and sell the right to color a pixel on a webpage. Read more
+[here](/GALLERY-README.md).
```
dckc: Please avoid _here_ links.

Instead, I suggest something like: Read [more about the gallery](/GALLERY-README.md)

GALLERY-README.md
```diff
@@ -6,43 +6,122 @@ number of additional features that showcase the unique affordances of
...
+thing are added and subtracted from each other, among other things.
+The assay is used, for example, when payments are deposited in purses,
+and we must calculate what the new balance of the purse is. It should
```
erights: You're asking the reader to hold too deep a stack of nested concepts in their head. Don't start a general explanation of assays while in the midst of explaining how a config has a slot, where a config is used to customize a makeMint. Explain first what the pixel dust looks like from a user of its issuer, purses, and payments. Then assays and amounts. Only then, how the pixel dust uses configures a makeMint to customize it to create these user-facing semantics.

README.md
```diff
@@ -1,13 +1,120 @@
...
+concepts:
+
+* __Vats__: All user code runs in what we call a `vat`. Within a vat,
+  code is run synchronously. Communication with another vat happens
+  asynchronously. The [SwingSet
+  infrastructure](https://github.com/Agoric/SwingSet) creates the vats
+  and makes communication between vats possible. 
+
+* __E() and infix bang (!)__: With our Swingset infrastructure, we can
+  talk to objects in another vat as if they were local, which is
```
dckc: as if they were local? that's over-selling, no? The difference isn't just syntax; infix-bang sends are eventual; i.e. they return promises.

https://en.wikipedia.org/wiki/Fallacies_of_distributed_computing

GALLERY-README.md
erights: I'm not saying you should use, specifically, the order I suggest above. It's just an example. Use your judgement.

erights: Given the overall level of this page, I think it is a distraction for it to explain about customizing mints at all. Just explain the pixels as an ERTP example in terms of the user-facing semantics. Explaining how one build new kinds of rights should come after being comfortable that there are many kinds of rights, and how to use them.

README.md
```diff
@@ -1,13 +1,120 @@
...
+ERTP.
+
+## Security Properties
+
+How does Alice know that she got paid real money? She could have been
+sent fake money, or she could have been sent money that was
+[double-spent](https://en.wikipedia.org/wiki/Double-spending). 
+
+When alice receives an alleged payment, she can call a method to know
+that the payment is 1) valid, and 2) exclusively hers:
```
erights: This, together with the following line of code, make it sound like `claimAll` is returning the same payment object, but somehow Alice now has exclusive access to it. It is not the `allegedPayment` that is now exclusively hers, it is the erights it contains, now transferred to a new `myExclusivePayment` object that is given only to her.

README.md
```diff
@@ -1,13 +1,120 @@
...
+  talk to objects in another vat as if they were local, which is
+  really cool! In order to do so, we need something more than the
+  "dot" syntax. So instead of `obj.foo()`, we either do `E(obj).foo()` or
+  the syntactic sugar, `obj!foo()`.
+
+* __Presences__: Presences are our name for the local object that
+  stands in for the remote object. So when we say `obj!foo()`, `obj`
+  is a presence because it represents a remote object that the message
+  `foo` will be passed to.
+
```
erights: > So when we say `obj!foo()`, `obj` is a presence because it represents a remote object that the message `foo` will be passed to.

This makes it sound like we only use the syntax above when `obj` is a presence. In fact, `obj` might just be a local object, a local promise, a local presence for a remote object, or a remote promise. The key thing is that all of these designate some actual object --- though it may be separated in space and time. The syntax means "deliver the message `foo()` to the actual object asynchronously, in its own turn, wherever and whenever it is, even if it is local."

GALLERY-README.md
katelynsills: Yup, agreed!

katelynsills: That is removed

README.md
erights: Perhaps introduce the syntax first with local objects, just as a way of invoking them asynchronously. Then, with that said, you can validly say that, using `E` or `!`, you can talk asynchronously to local and remote objects the same way, which is really cool!

README.md
katelynsills: Is this for search engine optimization?

README.md
katelynsills: Sure, Mark, introducing the syntax with local objects first sounds like a great solution. Dan, I'm not seeing any of fallacies mentioned in the wiki page as applicable here.

erights: The text Dan was reacting to sounds like its saying that you can use "!" to talk to objects remotely just as you use "." to talk to objects locally. Indeed, ***almost all historic RPC systems*** tried to do exactly that, blocking the sender at that point waiting for a round trip to complete, in order to emulate a synchronous function call. Given that network latencies are huge compared to local procedure calls, this was a terrible idea. The original Fallacies paper was written in response to such architectures. "Latency is zero".

By introducing "!" first as something different from ".", it is then true that remote "!" is like local "!" because the we designed the semantics of "!" (and indeed of non-blocking promises) so that it stayed within the restrictions of what could be transparently stretched over the network.

README.md
katelynsills: Good point. I added your language to this section and the E/! section. let me know if it still needs work!

README.md
katelynsills: Ah, gotcha. I think for modern JavaScript programmers, they will not necessarily assume that local objects are accessed synchronously.

README.md
erights: > So when we say `obj!foo()` and intend to send a message to a remote object, obj is a presence because it represents a remote object that the message foo will be passed to.

This phrasing is better but it still has the same problem. Maybe first state that `obj` is a presence of a remote object, and so serves as its local ambassador, representative, mascot, or something. (Don't say "proxy"!) Then explain that this enables us to send messages to the remote object by using "!" on the local presence of that remote object.

README.md
erights: Only once they understand that "!" is not just like ".", which the original text suggested.

README.md
katelynsills: Ok, I changed it to this

README.md
katelynsills: Sure, that could be made clearer

README.md
erights: Perfect.

README.md
dckc: I learned about it in the context of accessibility, but a quick search shows that yes, it's a search issue too:
[Poor Link Text Hurts Usability, Accessibility, and SEO](https://www.nngroup.com/articles/writing-links/)

katelynsills: Sounds like a good idea. Thanks for suggesting it. I'll make an issue and make those changes

PR #94, opened 2019-08-15
WIP: issuer operations preserve pipeline order

erights:
Use Promise.antiResolve from https://github.com/Agoric/eventual-send/pull/24 to keep issuer operations ordered without disrupting promise pipelining.

Would be a general technique for preserving pipelining over remote argument passing.

### DIFF
--- a/core/issuers.js
+++ b/core/issuers.js
@@ -83,23 +83,21 @@ Description must be truthy: ${description}`;
       return mint.mint(assay.empty(), name); // mint and issuer call each other
     },
 
-    claim(amount, srcPaymentP, name) {
-      return Promise.resolve(srcPaymentP).then(srcPayment => {
-        name = name !== undefined ? name : srcPayment.getName(); // use old name
-        return takePayment(srcPayment, paymentKeeper, amount, name);
-      });
+    claim(amount, srcPaymentR, name) {
+      const srcPayment = Promise.antiResolve(srcPaymentR);
+      name = name !== undefined ? name : srcPayment.getName(); // use old name
+      return takePayment(srcPayment, paymentKeeper, amount, name);
     },
 
-    claimAll(srcPaymentP, name) {
-      return Promise.resolve(srcPaymentP).then(srcPayment => {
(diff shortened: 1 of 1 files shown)


michaelfig:
This looks really good!  It's a useful optimisation and doesn't interfere with other naive uses of handled Promises.

Can you explain why you chose the `R` suffix in `srcPaymentR` (i.e. what it stands for)?

katelynsills:
Oh, this is interesting. What is anti-resolve?

michaelfig:
@erights' invention to be able to synchronously fetch the resolved value for a handled promise.  It returns `undefined` if the handled promise has not yet been resolved (or if it is not a handled promise).

I'm suggesting it be called `Promise.getPresence(myPromise)`.

erights:
> Can you explain why you chose the R suffix in srcPaymentR (i.e. what it stands for)?

Suggestion from @Chris-Hibbert 👍 

"R" for remote, but means "Not known to be local"
"P" for promise, but means, "Not known not to be a Promise and not known not to be remote"

This example itself demonstrates that, given `getPresence` (aka `antiResolve`), "R" does not imply that you do know it not to be a promise. It does imply that you know it to be one of:
   * A local object
   * A presence
   * A handled promise from which the non-thenable (local object or presence) is synchronously available.

"P" and "R" are not great for these. We need to bikeshed on our naming conventions.

michaelfig:
Should `antiResolve` return whatever was the fulfilment of the handler was, or only presences?  That may help indicate a good name.

erights:
> Should antiResolve return whatever was the fulfilment of the handler was, or only presences? That may help indicate a good name.

Whatever the fulfillment was. In the motivating case, payment arguments, the payments in question are the actual ones, not presences for them.

michaelfig:
Maybe `Promise.getResolution(somethingR)`, or `Promise.getFulfilment(somethingR)` (though the latter will definitely cause people to misspell it).

erights:
Obsolete

PR #95, opened 2019-08-15
gallery.js: setColor now subsumes changeColor, changeColorAll

michaelfig:
The first argument to setColor is a color.  The second argument
is an optional amount, which defaults to all pixels in the asset.

```js
home.gallery!tapFaucet()!getUse()!setColor('green')
```

This makes the demo a little more usable.  Next time we do a breaking
change of ERTP, we may want to remove the old changeColor* methods.

### DIFF
--- a/GALLERY-README.md
+++ b/GALLERY-README.md
@@ -83,7 +83,7 @@ with coloring the pixels in the underlying purse or payment) that can
 be called. To color, the user does:
 
 ```js
-E(useObj).changeColorAll('#000000');
+E(useObj).setColor('#000000');
 ```
 
 with the hex color of their choice. 
--- a/more/pixels/gallery.js
+++ b/more/pixels/gallery.js
@@ -101,9 +101,11 @@ export function makeGallery(
 
   function makeUseObj(issuer, asset) {
     const useObj = harden({
-      // change the color of the pixels in the amount after checking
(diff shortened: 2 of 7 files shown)


katelynsills:
One of the original designs for `changeColor` worked exactly like `setColor` in order to be simpler for users:
```
Story: the user changes the color of a pixel
The user calls tapFaucet() and receives a purse that contains one pixel
The user calls purse.changeColor(newColor, optionalAmount) to actually change the color
```
It was rejected in an ERTP design discussion by Mark and Dean because the default allowed for unintentional destructive behavior. If someone calls changeColor on their purse containing all their pixels, maybe they only meant to change the color of one pixel, and just forgot the amount. But now the user has to redo any design that they had. Mark and Dean preferred having `changeColor` and `changeColorAll` in order to throw an error if only a partial amount was intended, and that amount parameter was accidentally forgotten. 

I'm going to close this for now, but we can bring it up in the next design discussion.

michaelfig:
@dtribble actually suggested last night that we have this kind of API. 😝

katelynsills:
@dtribble !

katelynsills:
Ok, I talked to Dean and he says he realized this constraint and wants to keep the `changeColor`/`changeColorAll` distinction even if it's not as usable.

Issue #93, opened 2019-08-15
Gallery needs to hand out fresh "childPayments"


katelynsills:
There's a bug in the current implementation in which if the faucet is tapped repeated such that all pixels have been dispersed, and those pixel payments are deposited, and the faucet is tapped again, the payments from then on are empty. This happens because in the interest of conserving the ability to create new `childPayments`, calls to `getChildPayment` return the original childPayment if it already exists. Unfortunately, if that payment has been deposited or otherwise altered, the childPayment isn't created anew, but is returned with the empty or altered amount. 

See https://github.com/Agoric/ERTP/tree/test-revoked for a simple unit test for this.

See https://github.com/Agoric/ERTP/blob/4c73a80dd6f00a9780315071d08b70945848658f/more/pixels/pixelConfig.js#L67 for the line that creates the issue.

dtribble:
The singleton child pattern doesn't work: multiple requests for a child return the same payment, so if you transfer, revoke or otherwise consume it, then requesting "another" child will return the same, already-used payment. This is exactly what happens in tapFaucet.

There does not seem to be good motivation to allow the same child to be retrieved more than once. So instead, I propose that requesting a childPayment *always* revokes the prior child. That's more like the `claim` operation.

This would eliminate the revoke in [tapFaucet](https://github.com/Agoric/ERTP/blob/master/more/pixels/gallery.js#L189) and might eliminate the need for revokeChildren in general.

katelynsills:
Closed by Agoric/ERTP#96

warner:
in the old repo. this was ERTP issue 93

PR #96, opened 2019-08-15
Fix: revokeChildren() & getChildPayment() -> claimChild()

katelynsills:
There was a bug #93 in which, in the interest of conserving the childMint tokens, each call to `getChildPayment` returned the previously created payment if it was created. This PR fixes that bug and simplifies the process by combining `revokeChildren` and `getChildPayment` into `claimChild` as suggested by @dtribble. 

This PR adds tests that should catch this sort of bug in the future, and also cleans up some technical debt in how pixels are destroyed.


more/pixels/pixelMintKeeper.js
```diff
@@ -77,36 +76,27 @@ export function makePixelMintKeeper(assay) {
...
-      // amount must only contain a pixelList of length 1 for now
       const pixelList = assay.quantity(amount);
-      insist(
-        pixelList.length === 1,
-      )`amount must contain a pixelList of length 1 for now`;
+      pixelList.forEach(pixel => {
+        const strPixel = getString(pixel);
+        if (pixelToAsset.has(strPixel)) {
+          const asset = pixelToAsset.get(strPixel);
+          amount = assay.coerce(amount);
```
Chris-Hibbert: Please don't reassign to parameters.

more/pixels/pixelMintKeeper.js
```diff
@@ -77,36 +76,27 @@ export function makePixelMintKeeper(assay) {
...
-      const strPixel = getString(pixel);
-      insist(
-        pixelToAsset.has(strPixel),
-      )`pixel ${strPixel} could not be found to be destroyed`;
-      const asset = pixelToAsset.get(strPixel);
-      // amount is guaranteed to be there
-      amount = assay.coerce(amount);
+          const keeper = getKeeper(asset);
+          const originalAmount = keeper.getAmount(asset);
+          const newAmount = assay.without(originalAmount, amount);
```
Chris-Hibbert: The two amounts on the right are very different things. I think I'd try to distinguish between purseBalance and removePixelAmount. It might be okay to leave the second arg as amount if it started as allegedAmount or requestedAmount.

test/swingsetTests/gallery/vat-alice.js
```diff
@@ -369,6 +368,38 @@ function makeAliceMaker(E, log, contractHost) {
           showPaymentBalance('childPurseP', childPurseP);
           showPaymentBalance('grandchildPurseP', grandchildPurseP);
         },
+        async doGetAllPixels() {
+          log('++ alice.doGetAllPixels starting');
+          // No errors with this
+          // const pixels = [];
```
Chris-Hibbert: Does the comment ('No errors with this') reflect the current state or a goal? Is the commented out code intended to be reinstated? If so, this deserves a TODO. 

I can't tell whether the code used to pass incorrectly, would currently pass, but shouldn't , or is expected to pass in some future state. I guess it needs a more explicit description.

more/pixels/pixelMintKeeper.js
katelynsills: I think this is outside the scope of this PR since this PR didn't add these lines, and it's an ongoing discussion that we need to have. I would agree with you, though

test/swingsetTests/gallery/vat-alice.js
katelynsills: Good catch, I'll take that out

more/pixels/pixelMintKeeper.js
katelynsills: I think this code also wasn't changed in this PR, apart from putting it in an `if` block, but I would agree with you that `originalAmount` and `amount` are not very well distinguished. I can make the change you suggested in this PR.

test/swingsetTests/gallery/test-gallery.js
```diff
@@ -273,3 +273,37 @@ test('gallery - create childPayment, spend, revoke without SES', async t => {
   t.deepEquals(dump.log, expectedSpendAndRevoke);
   t.end();
 });
+
+const expectedGetAllPixels = [
```
dtribble: It's beyond the scope of this PR but we need a stronger unit test approach in general. These should be assertions in the tests.

katelynsills: I think you're right that the swingSetTests are inadequate. We have a separate unitTests folder that does allow us to have assertions in the tests. Those tests are included in this PR.

more/pixels/pixelConfig.js
```diff
@@ -110,29 +98,17 @@ function makePixelConfigMaker(
...
-                issuer,
-                superPurse.getBalance(),
-              );
-              childMint.revoke(childAmount);
-            }
+          // Revoke all descendants of this purse and mint a new purse
+          // from the child mint with the same quantity as the
+          // original purse
+          claimChild() {
+            getOrMakeChildMint(issuer);
```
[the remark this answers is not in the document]
katelynsills: I like that.

more/pixels/pixelConfig.js
```diff
@@ -110,29 +98,17 @@ function makePixelConfigMaker(
...
+            const childAmount = getChildAmount(issuer, superPurse.getBalance());
+            // Remove the amount of this payment from the purses and
+            // payments of the childMint. Removes recursively down the
+            // chain until it fails to find a childMint.
+            childMint.revoke(childAmount);
```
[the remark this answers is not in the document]
katelynsills: Good idea, I think I have to use the variable in scope, so I'll rename the function to `prepareChildMint`

more/pixels/pixelConfig.js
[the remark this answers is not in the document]
katelynsills: Added

PR #100, opened 2019-08-20
Extract Quantity Logic as a "Strategy" in Assays

katelynsills:
Our assays have a tremendous amount of duplicate code. Most of the difference in how an assay behaves can be reduced to the behavior of the set operations on quantities such as `empty()`, `with`, `without`, `includes`, etc. 

This PR extracts the set operation logic that acts on quantities and creates a configurable assay that takes in the set operation logic. Chris tells me that this pattern is known as a "strategy". 

* Assays are about Amounts, which are labeled quantities.
* Strategies are about quantity arithmetic.
* Assays use Strategies to do their quantity arithmetic, and then label the results, making new Amounts.


core/assay.js
```diff
@@ -0,0 +1,133 @@
...
+      },
+
+      equals(leftAmount, rightAmount) {
+        const leftQuantity = assay.quantity(leftAmount);
+        const rightQuantity = assay.quantity(rightAmount);
+        return strategy.equals(leftQuantity, rightQuantity);
+      },
+
+      // Set union of erights.
+      // Describe all the erights described by `leftAmount` and those
```
Chris-Hibbert: Combine the amounts described by 'leftAmount' and 'rightAmount'.

core/assay.js
```diff
@@ -0,0 +1,133 @@
...
+      // described by `rightAmount`.
+      with(leftAmount, rightAmount) {
+        const leftQuantity = assay.quantity(leftAmount);
+        const rightQuantity = assay.quantity(rightAmount);
+        return assay.make(strategy.with(leftQuantity, rightQuantity));
+      },
+
+      // Covering set subtraction of erights.
+      // If leftAmount does not include rightAmount, error.
+      // Describe the erights described by `leftAmount` and not described
```
Chris-Hibbert: The repetition of Describe/described doesn't help. How about something like "Return a representation of those rights included in 'leftAmount' but not included in 'rightAmount'."?

core/config/natAssay.js
```diff
@@ -0,0 +1,14 @@
+import harden from '@agoric/harden';
+
+import natStrategy from './natStrategy';
+import { makeAssayMaker } from '../assay';
+//
+// The default assay makes the default kind of amount.  The default
+// kind of amount is a labeled natural number describing a quantity of
+// fungible erights. The label describes what kinds of rights these
+// are. This is a form of labeled unit, as in unit typing.
```
Chris-Hibbert: Would you mind adding a line pointing out that we use natural numbers because floats have such problematic semantics? The strategy is to choose a small enough unit that it never has to deal with fractional parts. I didn't get that this was the point until I said something and got a reaction of disbelief that I had to interpret.

core/config/uniStrategy.js
```diff
@@ -0,0 +1,75 @@
...
+
+const makeUniStrategy = descriptionCoercer => {
+  const uniStrategy = harden({
+    insistKind: optDescription => {
+      if (optDescription === null) {
+        return null;
+      }
+      insist(
+        !!optDescription,
+      )`Uni optDescription must be either null or truthy ${optDescription}`;
```
Chris-Hibbert: add markers to demarcate ${optDescription}. quotes, angle brackets, parens, a prefix colon, or whatever.

more/pixels/pixelAssay.js
```diff
@@ -0,0 +1,17 @@
+import harden from '@agoric/harden';
+
+import { makePixelStrategy } from './pixelStrategy';
+import { makeAssayMaker } from '../../core/assay';
+
+// A pixelList is a naive collection of pixels in the form:
```
Chris-Hibbert: I created an issue to make a generic List Assay. (https://github.com/Agoric/ERTP/issues/101) I think it will be generally useful.

core/assay.js
```diff
@@ -0,0 +1,133 @@
...
+//
+// An assay is pass-by-presence, but is not designed to be usefully
+// passed. Rather, we expect each vat that needs to operate on amounts
+// will have its own local assay to do so.
+
+function makeAssayMaker(strategy) {
+  function makeAssay(label) {
+    mustBeComparable(label);
+
+    // memoize well formedness check of amounts
```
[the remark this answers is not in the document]
katelynsills: Good idea, I'll do that

core/config/uniStrategy.js
katelynsills: good idea

core/issuers.js
```diff
@@ -122,6 +122,10 @@ Description must be truthy: ${description}`;
       return assay;
     },
 
+    getStrategy() {
```
erights: Does anyone call this? Do you expect anyone to?

katelynsills:
> Is the following a good summary of this PR's central idea?
> 
> * Assays are about Amounts, which are labeled quantities.
> * Strategies are about quantity arithmetic.
> * Assays use Strategies to do their quantity arithmetic, and then label the results, making new Amounts.

Yes, that's right. I changed the description for this PR to include this and explain things better.

core/assay.js
```diff
@@ -0,0 +1,131 @@
...
+      getStrategy() {
+        return strategy;
+      },
+
+      // Given the raw quantity that this kind of amount would label, return
+      // an amount so labeling that quantity.
+      make(allegedQuantity) {
+        const amount = harden({
+          label,
+          quantity: strategy.insistKind(allegedQuantity),
```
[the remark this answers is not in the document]
katelynsills: I don't think this coerces. It doesn't for Nat (which throws an error) and it doesn't for pixelList. Is there somewhere that I've missed where coercion is happening?

core/issuers.js
katelynsills: Nothing calls it right now, but my state machine version of scooter plans to make use of this

core/assay.js
```diff
@@ -0,0 +1,131 @@
...
+      // received by pass-by-copy from an otherwise-identical remote
+      // amount? On success, return an amount object made by this
+      // assay. Otherwise error.
+      //
+      coerce(allegedAmount) {
+        if (brand.has(allegedAmount)) {
+          return allegedAmount;
+        }
+        if (!Object.prototype.hasOwnProperty.call(allegedAmount, 'quantity')) {
+          // This is not an amount. Let's see if it's a quantity. Will
```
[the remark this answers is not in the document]
katelynsills: This logic comes from the NatAssay, where we wanted users to be able to mint new tokens easily, such as calling `moolaMint.mint(1000)`. Without it, we would be forcing users to have to create an amount to pass into the `mint` method. Allowing users to pass in a quantity when minting makes the other kinds of mints easier to use as well. For instance, in the pixel gallery, the gallery currently does:

```
const amount = galleryPixelAssay.make(harden([pixel]));
const purse = galleryPixelMint.mint(amount);
```
This could be instead: 
```
const purse = galleryPixelMint.mint(harden[pixel]);
```
You pointed out that a user could be mistaken about what kind of mint they have, and try to mint type1 tokens but actually mint type2 tokens. This is a danger, but I think requiring the user to `make` the amount doesn't add much in solving this problem. By requiring the user to `make` the amount, all we are really checking is that the mint we hold matches the assay that we hold. If I am confused about my assay in the same way as I'm confused about my mint, it won't help me. We're not actually checking that the mint identification is correct. If we trust the user to keep track of what their assays mean, shouldn't we also trust them to keep track of which mints are which? Granted, it is a small double-check to make sure that your mint is the one associated with your assay, but it's a tradeoff and I'm not sure that it is helpful enough to make it more painful for the user to mint new tokens.

core/config/uniStrategy.js
```diff
@@ -0,0 +1,75 @@
...
+import { insist } from '../../util/insist';
+import {
+  sameStructure,
+  mustBeSameStructure,
+  mustBeComparable,
+} from '../../util/sameStructure';
+
+const makeUniStrategy = descriptionCoercer => {
+  const uniStrategy = harden({
+    insistKind: optDescription => {
```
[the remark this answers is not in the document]
katelynsills: The only use of `descriptionCoercer` comes from the `contractHost`:

```js
function descriptionCoercer(allegedDescription) {
  const seatDesc = seatDescriptions.get(allegedDescription.seatIdentity);
  mustBeSameStructure(seatDesc, allegedDescription);
  return seatDesc;
}

const makeUniAssayConfig = makeUniAssayConfigMaker(descriptionCoercer);
```

I interpret `descriptionCoercer` as additional, user-defined checks as part of `insistKind`. I don't think they can be easily unified.

core/config/uniAssay.js
```diff
@@ -0,0 +1,18 @@
+import harden from '@agoric/harden';
+
+import { makeUniStrategy } from './uniStrategy';
```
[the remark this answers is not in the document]
katelynsills: That's a really good point. I realized that I could take out the custom 'assays' entirely and have the `makeMint` config have a `strategy` slot rather than a `makeAssay` slot. Now there are only the strategies and the customizable `assay.js`

more/pixels/pixelStrategy.js
```diff
@@ -0,0 +1,30 @@
+import harden from '@agoric/harden';
+
+import {
+  insistPixelList,
+  includesPixelList,
+  withPixelList,
+  withoutPixelList,
+} from './types/pixelList';
```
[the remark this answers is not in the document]
katelynsills: I have a lot of tests associated with the separate pixelList module that I would have to edit slightly to use the strategy instead. I think you're right that they should be combined. Would it be ok to add an issue for that and do so in a separate PR?

core/assay.js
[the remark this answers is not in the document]
erights: Since you're code passes the `allegedQuantity` in to this method and uses what this method returns, you are enabling this method to coerce. Although we don't have an example, this was the original design intention.

If we wish this method only to approve or reject, we should not give it the opportunity to coerce. We should write it as

```js
make(allegedQuantity) {
  strategy.insistKind(allegedQuantity);
  const amount = harden({
    label,
    quantity: allegedQuantity,
```

We should either change its name to something that suggests coercion, or we should change the code that calls it so that it isn't able to coerce.

I think having it be able to coerce is better in this case.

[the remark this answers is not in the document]
katelynsills: I wanted to better understand your use of the word 'coerce'. My understanding of 'coerce' is that it is implicit type conversion from typeA to typeB. That doesn't seem to be what is happening here. For instance, if I call `make(3.9)` and `strategy.insistKind` is `Nat`, `make` will error rather than try to make `3.9` into the Nat `3`. By my definition, there is no coercion.

I think you are saying that because `insistKind` returns the `allegedQuantity` if it doesn't error (it is already a Nat), it is doing coercion. Is that right? I don't see how this fits the definition of coercion since no type conversion occurs. 

It also sounds like you might be saying that anything with the name `insist` should not be returning anything. Is that right? If that's the case, would you be ok with calling `insistKind` `insistKindAndReturn` or something similar? It is really doing an insist, but it is also returning a value but only if there is no error.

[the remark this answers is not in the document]
Chris-Hibbert: 'coerce' here is a bit more general than only type conversion. The point is that since it returns an object, it can modify the object in arbitrary ways. (The assay is the authority on what objects are valid, so the issue isn't that we're vulnerable to it.) Like with the design of the peg, it might substitute an amount with a presence for an issuer for an equivalent amount with a local issuer. 

This is partly about terminology. A function called insistFoo() should return true or false. A function that takes an arg and returns something represents the same thing is more appropriately called coerceFoo().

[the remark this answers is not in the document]
katelynsills: Gotcha. I think that the proper usage of `insistFoo` (following the behavior of the`insist` util) is not that it returns true or false but that it throws or doesn't throw. I just pushed up a change that moves the `insistKind` call up so that it acts more like a guard. The `quantity` is then just set to `allegedQuantity` as Mark had above.

core/assay.js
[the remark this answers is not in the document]
erights: In your example

> `const purse = galleryPixelMint.mint(harden[pixel]);`

First, did you mean

> `const purse = galleryPixelMint.mint(harden([pixel]));`

Second, what kind of thing is `pixel`?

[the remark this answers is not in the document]
erights: Nevermind. I was misreading your code and intent! I missed the `!` in the condition and thought you were doing something completely different whose motivation I could see but didn't like. Please ignore my comments above.

What if it has a 'quantity' field but no 'label' field? This kind of duck type sniffing is inherently dangerous. But now I understand the real motivation and I do not object. Resolving.

core/config/strategies/natStrategy.js
```diff
@@ -12,8 +10,15 @@ import { makeAssayMaker } from '../assay';
...
+  empty: _ => 0,
+  isEmpty: nat => nat === 0,
+  includes: (whole, part) => whole >= part,
+  equals: (left, right) => left === right,
+  with: (left, right) => Nat(left + right),
+  without: (whole, part) => Nat(whole - part),
+});
 
-export { makeNatAssay };
+export default natStrategy;
```
erights: Why not 

```js
export { natStrategy };
```

?

core/assay.js
```diff
@@ -21,111 +21,108 @@ import { mustBeSameStructure, mustBeComparable } from '../util/sameStructure';
...
-        });
-        brand.add(amount);
-        return amount;
-      },
-
-      // Is this an amount object made by this assay? If so, return
-      // it. Otherwise error.
-      vouch(amount) {
-        insist(brand.has(amount))`\
+function makeAssay(label, strategy) {
```
erights: Good! Glad to see an unnecessary level get collapsed away.

more/pixels/pixelStrategy.js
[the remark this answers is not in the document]
erights: > Would it be ok to add an issue for that and do so in a separate PR?

Yes. Resolving.

core/assay.js
```diff
@@ -0,0 +1,129 @@
...
+    },
+
+    getStrategy() {
+      return strategy;
+    },
+
+    // Given the raw quantity that this kind of amount would label, return
+    // an amount so labeling that quantity.
+    make(allegedQuantity) {
+      strategy.insistKind(allegedQuantity);
```
erights: This is now self consistent, so LGTM for this PR. However, I think the opposite consistent stance, that the strategy contribute a quantity coercing behavior, may be preferable. Since we have no current need for such a strategy-based coercion, we can mull this over non-urgently.

core/config/strategies/natStrategy.js
katelynsills: Good idea, I will do that.

katelynsills:
> My biggest previous criticism was due to a misreading and misunderstanding on my part. (I missed a `!` on a condition.) Thanks for patiently clarifying.

Thank you for the thoughtful review!

core/assay.js
```diff
@@ -0,0 +1,129 @@
...
+
+  // The brand represents recognition of the amount as authorized.
+  const brand = new WeakSet();
+
+  const assay = harden({
+    getLabel() {
+      return label;
+    },
+
+    getStrategy() {
```
Chris-Hibbert: Now that this is mostly settled, would you update the .chainmail file? 

I think my attitude is that the .chainmail file should focus on the customer API rather than internal details. `getStrategy()` is visible to the API, so it should at least appear in that file. 

OTOH, strategies are a crucial part of the extension story, so the interface they present and while is relied on by other parts of the code needs to be written up somewhere. That could either be in one of the .js files (assays? issuers? I'm not sure) or in the .chainmail file if it makes more sense to segregate all the documentation there.

katelynsills: Sure, that sounds good. I made an issue for this

PR #122, opened 2019-09-16
Add Seat Mint

katelynsills:
This PR is part of the current Zoe implementation. (I've [split Zoe up for better review](https://github.com/Agoric/ERTP/issues/134).) It gives users the ability to easily create ERTP payments for "seats" (we need to sort out the various names for ERTP, but let's do that in one big meeting rather than in this PR.) for their contracts.  A seat can either represent the right to make an offer (an "invite") or the right to claim the outcome of making an offer (a "position"). Seats are ERTP payments which have a use object that can be obtained by calling `unwrap`. `Unwrap` burns the underlying payment (or in the case of a purse, withdraws the full balance and burns that, leaving the underlying purse empty). The interface of the use object is entirely flexible and up to the developer. 

Two modifications were necessary:
1. Updating the "trait-like" customization of `issuers.js` with an approach invented by @erights  that gives the custom code access to the full payment, purse, mint, issuer while still not allowing the custom code the ability to override the core ERTP methods. This turned out to be necessary to be able to burn the payments within the custom code. When the custom code only had access to the core/super payment, purse, etc., the attempt to burn would fail because of course, the core payment, purse, etc. were different objects than the final version (after customization) that was actually stored in the weakmap.
2. Making UniStrategy more generalizable to be able to reuse it. I also cleaned up the language a bit - please let me know if I've left anything out.

### DIFF
--- a/core/config/configs.md
+++ b/core/config/configs.md
@@ -1,0 +1,152 @@
+## Custom Mints and Other Configurations
+
+The `makeMint` function in `issuers.js` takes in a configuration
+function that can change a number of things about how mints, issuers,
+purses, and payments are made. 
+
+By default, `makeBasicFungibleConfig` is used. This creates a mint for
+a fungible token with no custom methods.
+
+But, imagine that we want to add a custom method to the `mint` object.
+Maybe we want to be able to know how much has been minted so far. To
+do so, we will need to create our own custom configuration. 
+
+Our custom configuration will look like this:
+
+```js
+const makeTotalSupplyConfig = () => {
+
+  function* makePaymentTrait(_superPayment) {
+    yield harden({});
+  }
+
+  function* makePurseTrait(_superPurse) {
+    yield harden({});
+  }
+
+  function* makeMintTrait(_superMint, _issuer, _assay, mintKeeper) {
+    return yield harden({
+      getTotalSupply: () => mintKeeper.getTotalSupply(),
+    });
+  }
+
+  function* makeIssuerTrait(_superIssuer) {
+    yield harden({});
+  }
+
+  return harden({
+    makePaymentTrait,
+    makePurseTrait,
+    makeMintTrait,
+    makeIssuerTrait,
+    makeMintKeeper: makeTotalSupplyMintKeeper,
+    strategy: natStrategy,
+    makeMintTrait,
+  });
+};
+```
+
+In this custom configuration, we've done two things: we've added a
+method to `mint` called `getTotalSupply`, and we've changed the
+`mintKeeper` to a custom mintKeeper that keeps track of the total
+supply for us (more on this in separate documentation). 
+
+Let's take a look into how we are able to add new methods to mints,
+issuers, purses, and payments. 
+
+In `makePaymentTrait`, we take `superPayment` as a parameter. The
+`superPayment` is constructed in `issuers.js` and has all of the
+methods we are familiar with:
+
+```js
+const superPayment = harden({
+  getIssuer() {
+    return issuer;
+  },
+  getBalance() {
+    return paymentKeeper.getAmount(payment);
+  },
+  getName() {
+    return name;
+  },
+});
+```
+
+Our custom code, `makePaymentTrait`, returns an object with custom methods.
+These methods will be added to the `superPayment`. For this particular
+customization, we want to leave payments alone, so we will create a
+generator function that yields an empty object:
+
+```js
+function* makePaymentTrait(_superPayment) {
+  yield harden({});
+}
+```
+
+However, we do want to add an extra method to mints. So,
+`makeMintTrait` is defined as:
+
+```js
+function* makeMintTrait(_superMint, _issuer, _assay, mintKeeper) {
+  return yield harden({
+    getTotalSupply: () => mintKeeper.getTotalSupply(),
+  });
+}
+```
+
+Back in `issuers.js`, our custom methods will be combined with the
+"core" methods already present. Here's how that works for our custom
+mint methods:
+
+```js
+const makeMintTraitIter = makeMintTrait(coreMint, issuer, assay, mintKeeper);
+const mintTrait = makeMintTraitIter.next().value;
+const mint = harden({
+  ...mintTrait,
+  ...coreMint,
+});
+makeMintTraitIter.next(mint);
+```
+
+## The Trait Pattern
+
+As a general rule, we do not want the customization code to be able to
+override the methods already on the mints, issuers, purses or
+payments. It would hardly be a standard if the standard methods acted
+entirely differently! On the other hand, we do know that there is a
+genuine need to have slightly different purposes expressed. 
+
+This is why we've chosen to use the [Trait
+pattern](https://traitsjs.github.io/traits.js-website/), in which
+custom behavior can be defined and recombined easily. (Note: We don't
+currently use the Trait.js infrastructure, but the philosophy is the
+same.) Furthermore, if we always combine the methods such that the
+"core" methods are last, they cannot be overridden by the custom
+methods. 
+
+So we're good, right? Well, almost.
+
+## Why generator functions?
+
+Sometimes, we want to burn the payment in the custom method. That
+caused a problem, because without generators, we don't actually have
+access to the payment! We only have access to the `superPayment` that
+is passed in as a parameter. When we call `issuer.burnAll(payment)`,
+the payment gets looked up in a WeakMap of all the payments and their
+current balances. However, the superPayment won't be in that WeakMap.
+The finished payment (the superPayment plus the custom methods) is
+what is in the WeakMap. So we must have access to *that* payment. 
+
+There's a contradiction there. We wanted to ensure that the custom
+code couldn't override or mess with the core methods, but now we're
+saying that we should hand off the whole, un-`hardened` payment to the
+custom code. 
+
+There's a way to get around this contradiction: generator functions.
+We can still use the trait pattern. However, we can also give access
+to the `hardened`, final version of the payment that is the core + custom methods.
+
+Here's how that happens:
+
+* temporal dead zone
+* ??
--- a/core/config/configs.md
+++ b/core/config/configs.md
@@ -127,12 +127,13 @@ So we're good, right? Well, almost.
 
 Sometimes, we want to burn the payment in the custom method. That
 caused a problem because with normal functions we don't actually have
-access to the payment! We only have access to the `corePayment` that
+access to the full payment, the one which is built by adding methods to the
+`corePayment`. We only have access to the `corePayment` that
 is passed in as a parameter. When we call `issuer.burnAll(payment)`,
 the payment gets looked up in a WeakMap of all the payments and their
 current balances. However, the corePayment won't be in that WeakMap.
-The finished payment (the corePayment plus the custom methods) is
-what is in the WeakMap. So we must have access to *that* payment. 
+The finished payment (the corePayment plus the custom methods) is what
+is in the WeakMap. So we must have access to *that* payment. 
 
 There's a contradiction there. We wanted to ensure that the custom
 code couldn't override or mess with the core methods, but now we're
--- a/core/config/configs.md
+++ b/core/config/configs.md
@@ -16,21 +16,21 @@ Our custom configuration will look like this:
 ```js
 const makeTotalSupplyConfig = () => {
 
-  function* makePaymentTrait(_superPayment) {
+  function* makePaymentTrait(_corePayment) {
     yield harden({});
   }
 
-  function* makePurseTrait(_superPurse) {
+  function* makePurseTrait(_corePurse) {
     yield harden({});
   }
 
-  function* makeMintTrait(_superMint, _issuer, _assay, mintKeeper) {
+  function* makeMintTrait(_coreMint, _issuer, _assay, mintKeeper) {
     return yield harden({
       getTotalSupply: () => mintKeeper.getTotalSupply(),
     });
   }
 
-  function* makeIssuerTrait(_superIssuer) {
+  function* makeIssuerTrait(_coreIssuer) {
     yield harden({});
   }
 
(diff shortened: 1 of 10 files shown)


core/config/uniAssayConfig.js
```diff
@@ -4,15 +4,29 @@ import { noCustomization } from './noCustomization';
...
 import { makeUniStrategy } from './strategies/uniStrategy';
 
+import { insist } from '../../util/insist';
+import { mustBeComparable } from '../../util/sameStructure';
+
 // This UniAssay config is used to create invites and similar assets.
 // It does not customize the purses, payments, mints, or issuers, and
 // it uses the core mintKeeper as well as the uniAssay (naturally)
-function makeUniAssayConfigMaker(descriptionCoercer) {
+const makeCustomInsistKind = descriptionCoercer => optDescription => {
```
Chris-Hibbert: This syntax 

```
const makeCustomFn = foo => bar => {
   ... bar ... foo ...
}

function n(foo) {
  const customFn = makeCustomFn(foo);
  ... customFn ...
}
```

looks convoluted to me, though I was able to understand it.  Would it better as

```
function n(foo) {
  function customFn(bar) {
   ... bar ... foo ...
};
  ... customFn ...
}
```

I guess this comes down to two style questions. 

1) How recognizable is ```foo => bar => { ... }``` as syntax for currying to javascript developers? It seemed weird but convenient to me, but I don't know what's common.

2) Is it helpful to write a currying function even though the value to be curried is immediately available? In this case, it does have the advantage of moving 8 lines of code outside a function definition.

core/config/seatConfig.js
```diff
@@ -0,0 +1,54 @@
...
+ * purses
+ */
+function makeSeatConfigMaker(makeUseObjForPayment, makeUseObjForPurse) {
+  function makeSeatConfig() {
+    return harden({
+      makeCustomPayment(superPayment, issuer) {
+        const payment = harden({
+          ...superPayment,
+          // This creates a new use object which destroys the payment
+          unwrap: () => makeUseObjForPayment(issuer, payment),
```
Chris-Hibbert: Why was the switch from makePTrait to makeCustomP helpful? It looks like the traits also had access to the issuer and payment when they were necessary. Couldn't the payment have been destroyed in that framework?

The improvement isn't obvious, and it has the drawback of allowing overriding of methods from the core set.

core/seatMint.js
```diff
@@ -0,0 +1,62 @@
...
+ * objects) to use objects
+ */
+const makeSeatMint = (description = 'seats') => {
+  const idObjsToSeats = new WeakMap();
+
+  const addUseObj = (idObj, useObj) => {
+    idObjsToSeats.set(idObj, useObj);
+  };
+
+  const makeUseObj = seatQuantity => {
```
Chris-Hibbert: It's surprising that access to the quantity gives access to the use Object. In our other contracts, the quantity is used as part of the description of what's being exchanged. wouldn't we want to be able to share the quantity of an invitation in order to discuss what the invitation is for?

test/unitTests/core/test-seat.js
```diff
@@ -0,0 +1,135 @@
...
+ * or:
+ *
+ * {
+ *   id: {},
+ *   offerMade: [rule1, rule2],
+ * }
+ *
+ */
+
+const allTrue = (prev, curr) => prev && curr;
```
Chris-Hibbert: call it bothTrue?

test/unitTests/core/test-seat.js
```diff
@@ -0,0 +1,135 @@
...
+        claim: () => {
+          return [];
+        },
+      });
+    }
+    return harden({});
+  };
+
+  const purse1Quantity = harden({
+    id: harden({}),
```
Chris-Hibbert: does the harden here add anything, since it's wrapped in an outer harden?

core/config/seatMint.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+const makeSeatMint = () => {
+  const idObjsToSeats = new WeakMap();
+
+  const addUseObj = (idObj, useObj) => {
+    idObjsToSeats.set(idObj, useObj);
+  };
+
+  const makeUseObj = seatQuantity => {
+    return harden(idObjsToSeats.get(seatQuantity.id));
```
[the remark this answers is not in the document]
katelynsills: No, you can't get access to the use object by having access to the seatQuantity. 

The method `makeUseObj` and the `idObjsToSeats` weakmap are closely held and don't leave this file. If you look at the bottom of the file, you can see that only `seatMint`, `seatIssuer` and `addUseObj` are exposed to the code that called `makeSeatMint`, so there is no way that 1) the holder of the makeMint authority can get use objects, let alone someone else.

core/config/seatConfig.js
katelynsills: I think there's a misunderstanding here. The `super` or `core` object was passed in for `makePTrait`, but the custom code there had no access to the full object (super/core + custom methods). By moving to `makeCustomP`, the custom code does have access to the full object, because they are creating it and returning it.

Happy to talk about this more, but to be clear, this wasn't an "improvement" - it was necessary to be able to burn the underlying payment in the custom code.

core/seatMint.js
katelynsills: I think there's a misunderstanding here. Access to the quantity doesn't give access to the use object, so we can share the quantity freely.

test/unitTests/core/test-seat.js
katelynsills: It doesn't add anything technically, but I like hardening whenever I create a new object so that I don't forget it when it's actually important.

test/unitTests/core/test-seat.js
katelynsills: Do you think that's clearer as the parameter to `array.reduce`? To me it's clearer that I'm checking that everything in the array is true when it's called `allTrue` instead of `bothTrue`, even though of course reduce is comparing two items: a new item to the reduction of everything before it.

core/config/uniAssayConfig.js
katelynsills: Hmm, I don't know if I can answer how recognizable it is. A [JavaScript book](https://leanpub.com/javascriptallongesix/read) that I really like (and has been on the front page of Hacker News a couple of times) uses it a lot. The functional programming community uses it a lot. I think it adds more clarity than having `makers` for `makers`, but that could be just me.

Re: 2: I could have put that whole function inside `makeUniAssayConfigMaker` as you say, relying on `descriptionCoercer` to be in scope. I tend not to like this kind of entanglement and reliance on scope because it makes it harder to test and refactor, IMO. What do you think?

Chris-Hibbert: re: #1. Thanks for the reference. I'm happy to use this as justification for what JS programmers should be familiar with. I'll try to find time to read more of it so I can have an opinion on whether we should recommend it as a style guide in general.

re: #2. I was about to react to the "harder to test" part of your response, and say that this would be a strong argument when used to actually write tests, but then I noticed "and refactor". I now agree that this is clearer and easier to manipulate, even though the variable is referenced both abstractly and concretely. Having the definition and use so close to one another also arguably makes it easier to see what's going on.

test/unitTests/core/test-seat.js
Chris-Hibbert: `allTrue()` makes it seem like a predicate I could pass more than two parameters to. This would result in it ignoring all but the first two, which could lead to some surprising results. If it could only be called in `reduce()`, this wouldn't be an issue, but I don't see anything that prevents other uses.

I do agree that it reads better in `reduce()` calls.

core/seatMint.js
Chris-Hibbert: Ah, I see. makeUseObj() is an internal function, so no one external can use it. Thanks for the correction.

test/unitTests/core/test-seat.js
katelynsills: Ok, that makes sense to me. I just changed `allTrue` to `bothTrue`

Chris-Hibbert:
Please add a description of the "bizarre" "weird way to compose traits". The definer of new traits is now required to write them as generators. 

I'd like to see an explanation of how this helps protect the issuer, and what the implementor of a new configuration can do in the generators.

Have you made any progress on "invent[ing] a reusable abstraction for the boilerplate"?

core/config/strategies/uniStrategy.js
```diff
@@ -7,64 +7,58 @@ import {
...
-const makeUniStrategy = (descriptionCoercer = d => d) => {
+// The uniStrategy represents quantities that can never be combined.
+// For example, usually there is only one invite in an invite purse or
+// payment. (Using a listStrategy is an alternative, but when there is
+// usually a quantity of one, it's bothersome to always have to grab
+// the first item in the list rather than just represent the item
+// itself.)
+
+// The uni quantities are either empty (null) or unique. (The unique
+// quantities will often include an empty object for identity).
```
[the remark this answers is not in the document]
katelynsills: Sure, I'll edit it. I meant that if a uni quantity is fulfilling the role of an id, we prefer using an empty, unique object.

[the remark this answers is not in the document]
katelynsills: Edited

katelynsills:
> Please add a description of the "bizarre" "weird way to compose traits". The definer of new traits is now required to write them as generators.
> 
> I'd like to see an explanation of how this helps protect the issuer, and what the implementor of a new configuration can do in the generators.
> 
> Have you made any progress on "invent[ing] a reusable abstraction for the boilerplate"?

I added more documentation on this, but I think only Mark knows the details. In `configs.md`, I started off an explainer and left room for Mark to do his magic:

```
There's a way to get around this contradiction: generator functions.
We can still use the trait pattern. However, we can also give access
to the `hardened`, final version of the payment that is the core + custom methods.

Here's how that happens:

* temporal dead zone
* ??
```

core/config/configs.md
```diff
@@ -0,0 +1,152 @@
...
+## The Trait Pattern
+
+As a general rule, we do not want the customization code to be able to
+override the methods already on the mints, issuers, purses or
+payments. It would hardly be a standard if the standard methods acted
+entirely differently! On the other hand, we do know that there is a
+genuine need to have slightly different purposes expressed. 
+
+This is why we've chosen to use the [Trait
+pattern](https://traitsjs.github.io/traits.js-website/), in which
```
Chris-Hibbert: I think this is a [better link for traits](https://en.wikipedia.org/wiki/Trait_(computer_programming)). It describes the OO pattern rather than a JS package we're not using.

core/config/configs.md
```diff
@@ -0,0 +1,152 @@
...
+  getName() {
+    return name;
+  },
+});
+```
+
+Our custom code, `makePaymentTrait`, returns an object with custom methods.
+These methods will be added to the `superPayment`. For this particular
+customization, we want to leave payments alone, so we will create a
+generator function that yields an empty object:
```
Chris-Hibbert: I like starting with a description of the no-change case.

core/config/configs.md
```diff
@@ -0,0 +1,152 @@
...
+  ...coreMint,
+});
+makeMintTraitIter.next(mint);
+```
+
+## The Trait Pattern
+
+As a general rule, we do not want the customization code to be able to
+override the methods already on the mints, issuers, purses or
+payments. It would hardly be a standard if the standard methods acted
```
Chris-Hibbert: I think this starts too softly ('as a general rule'). We're actually intending to prohibit changes to the behavior  of the core methods, aren't we? And we're not defining actual 'standard's.

What do you think of this:

We want the core behavior of mints, issuers, purses and payments to be consistent and reliable across different types of assets. On the other hand, we do know that there is a
genuine need to have slightly different purposes expressed.

core/config/configs.md
```diff
@@ -0,0 +1,152 @@
...
+same.) Furthermore, if we always combine the methods such that the
+"core" methods are last, they cannot be overridden by the custom
+methods. 
+
+So we're good, right? Well, almost.
+
+## Why generator functions?
+
+Sometimes, we want to burn the payment in the custom method. That
+caused a problem, because without generators, we don't actually have
```
Chris-Hibbert: 'without generators' presumes the solution. How about 'with normal functions'?

core/config/configs.md
```diff
@@ -0,0 +1,152 @@
...
+access to the payment! We only have access to the `superPayment` that
```
Chris-Hibbert: I think the connection between the payment and 'superPayment' isn't obvious from the fragments cited here. [Isn't it `corePayment` in the code? That would make it clearer since the coreMint construction is quoted above.]

I'd say something like 'we don't actually have access to the payment (which is built by adding methods to the superPayment's methods)'

core/config/configs.md
```diff
@@ -0,0 +1,152 @@
...
+current balances. However, the superPayment won't be in that WeakMap.
+The finished payment (the superPayment plus the custom methods) is
+what is in the WeakMap. So we must have access to *that* payment. 
+
+There's a contradiction there. We wanted to ensure that the custom
+code couldn't override or mess with the core methods, but now we're
+saying that we should hand off the whole, un-`hardened` payment to the
+custom code. 
+
+There's a way to get around this contradiction: generator functions.
```
Chris-Hibbert: this revelation works better if we don't telegraph the answer above, I think.

core/config/configs.md
```diff
@@ -0,0 +1,152 @@
...
+We can still use the trait pattern. However, we can also give access
+to the `hardened`, final version of the payment that is the core + custom methods.
+
+Here's how that happens:
+
+* temporal dead zone
+* ??
```
Chris-Hibbert: more details needed here.  :-)

I like the approach even though we're waiting for help filling in the details.

core/config/configs.md
katelynsills: Sounds good!

core/config/configs.md
katelynsills: Yeah, good point. In the code, we use both "super" and "core" to mean the same thing. I've renamed the uses of "super" to "core" so that it's uniform.

katelynsills: Done

core/issuers.js
```diff
@@ -67,10 +67,13 @@ Description must be truthy: ${description}`;
 
     // makePaymentTrait is defined in the passed-in configuration and adds
     // additional methods to corePayment
+    const makePaymentTraitIter = makePaymentTrait(corePayment, issuer);
+    const paymentTrait = makePaymentTraitIter.next().value;
     const payment = harden({
-      ...makePaymentTrait(corePayment, issuer),
+      ...paymentTrait,
```
Chris-Hibbert: Should the traits be plural?  The `...` makes me think there might be more than one.

katelynsills: I could be wrong but my understanding of `trait` was that a trait could encompass multiple methods.

katelynsills:
Closing because we decided to squash the Zoe PRs

PR #123, opened 2019-09-16
Add "areRightsConserved" and "isOfferSafe" functions with tests

katelynsills:
This PR is part of the Zoe implementation. It adds helper functions for determining whether a reallocation of quantities conserves rights (has the same total quantity per issuer) and whether a reallocation is "offer-safe" for each user.

### DIFF
--- a/core/zoe/areRightsConserved.js
+++ b/core/zoe/areRightsConserved.js
@@ -1,0 +1,47 @@
+import { bothTrue, transpose } from './utils';
+
+/**
+ * The columns in a `quantities` matrix are per issuer, and the rows
+ * are per player. We want to transpose the matrix such that each
+ * row is per issuer so we can do 'with' on the array to get a total
+ * per issuer and make sure the rights are conserved.
+ * @param  {strategy[]} strategies - an array of strategies per issuer
+ * @param  {quantity[]} quantities - an array of arrays with a row per
+ * player indexed by issuer
+ */
+const sumByIssuer = (strategies, quantities) =>
+  transpose(quantities).map((quantitiesPerIssuer, i) => {
+    return quantitiesPerIssuer.reduce(strategies[i].with);
+  });
+
+/**
+ * Does the left array of summed quantities equal the right array of
+ * summed quantities?
+ * @param  {strategy[]} strategies - an array of strategies per issuer
+ * @param  {quantity[]} leftQuantities - an array of total quantities per issuer
+ * @param  {quantity[]} rightQuantities - an array of total quantities per issuer
+ * indexed by issuer
+ */
+const isEqualPerIssuer = (strategies, leftQuantities, rightQuantities) =>
+  leftQuantities
+    .map((leftQ, i) => strategies[i].equals(leftQ, rightQuantities[i]))
+    .reduce(bothTrue, true);
+
+/**
+ * `areRightsConserved` checks that the total quantity per issuer stays
+ * the same regardless of the reallocation.
+ * @param  {strategy[]} strategies - an array of strategies per issuer
+ * @param  {quantity[][]} previousQuantities - array of arrays where a row
+ * is the array of quantities for a particular player, per
+ * issuer
+ * @param  {quantity[][]} newQuantities - array of arrays where a row
+ * is the array of reallocated quantities for a particular player, per
+ * issuer
+ */
+function areRightsConserved(strategies, prevQuantities, newQuantities) {
+  const sumsPrevQuantities = sumByIssuer(strategies, prevQuantities);
+  const sumsNewQuantities = sumByIssuer(strategies, newQuantities);
+  return isEqualPerIssuer(strategies, sumsPrevQuantities, sumsNewQuantities);
+}
+
+export { areRightsConserved };
--- a/core/zoe/isOfferSafe.js
+++ b/core/zoe/isOfferSafe.js
@@ -1,0 +1,119 @@
+import { bothTrue } from './utils';
+import { insist } from '../../util/insist';
+
+/**
+ * `isOfferSafeForPlayer` checks offer-safety for a single player.
+ *
+ * Note: This implementation checks whether we refund for all rules or
+ * return winnings for all rules. It does not allow some refunds and
+ * some winnings, which is what would happen if you checked the rules
+ * independently. It *does* allow for returning a full refund plus
+ * full winnings.
+ *
+ * @param  {strategy[]} strategies - an array of strategies ordered in
+ * the same order as the corresponding issuers
+ * @param  {offerDescElem[]} offerDesc - the offer description, an
+ * array of objects that have a rule and amount, in the same order as
+ * the corresponding issuers. The offer description is a player's
+ * understanding of the contract that they are entering when they make
+ * an offer. OfferDescElements are structured in the form `{ rule:
+ * descriptionString, amount}`
+ * @param  {quantity[]} quantities - an array of quantities ordered in
+ * the same order as the corresponding issuers. This array of quantities
+ * is the reallocation to be given to a player.
+ */
+function isOfferSafeForPlayer(strategies, offerDesc, quantities) {
+  insist(
+    strategies.length === offerDesc.length &&
+      strategies.length === quantities.length,
+  )`strategies, the offer description, and quantities must be arrays of the same length`;
+
+  const allowedRules = [
+    'offerExactly',
+    'offerAtMost',
+    'wantExactly',
+    'wantAtLeast',
+  ];
+
+  // If we are refunding the player, are their allocated amounts
+  // greater than or equal to what they said they had at the beginning?
+  const refundOk = offerDesc
+    .map((offerDescElem, i) => {
+      if (offerDescElem === null || offerDescElem === undefined) {
+        return true;
+      }
+      insist(
+        allowedRules.includes(offerDescElem.rule),
+      )`The rule ${offerDescElem.rule} was not recognized`;
+      // If the rule was 'offerExactly', we should make sure that the
+      // user gets it back exactly in a refund. If the rule is
+      // 'offerAtMost' we need to ensure that the user gets back the
+      // amount or greater. If the rule is something else, anything
+      // we give back is fine.
+      if (offerDescElem.rule === 'offerExactly') {
+        return strategies[i].equals(
+          quantities[i],
+          offerDescElem.amount.quantity,
+        );
+      }
+      if (offerDesc.rule === 'offerAtMost') {
+        return strategies[i].includes(
+          quantities[i],
+          offerDescElem.amount.quantity,
+        );
+      }
+      return true;
+    })
+    .reduce(bothTrue, true);
+
+  // If we are not refunding the player, are their allocated amounts
+  // greater than or equal to what they said they wanted at the beginning?
+  const winningsOk = offerDesc
+    .map((offerDescElem, i) => {
+      if (offerDescElem === null || offerDescElem === undefined) {
+        return true;
+      }
+      insist(
+        allowedRules.includes(offerDescElem.rule),
+      )`The rule ${offerDescElem.rule} was not recognized`;
+      // If the rule was 'wantExactly', we should make sure that the
+      // user gets exactly the amount specified in their winnings. If
+      // the rule is 'wantAtLeast', we need to ensure that the user
+      // gets back winnings that are equal or greater to the amount.
+      // If the rule is something else, anything we give back is fine.
+      if (offerDescElem.rule === 'wantExactly') {
+        return strategies[i].equals(
+          quantities[i],
+          offerDescElem.amount.quantity,
+        );
+      }
+      if (offerDescElem.rule === 'wantAtLeast') {
+        return strategies[i].includes(
+          quantities[i],
+          offerDescElem.amount.quantity,
+        );
+      }
+      return true;
+    })
+    .reduce(bothTrue, true);
+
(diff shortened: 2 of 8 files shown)


core/zoe/areRightsConserved.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param  {array} rightQuantities - an array of total quantities per issuer
+ * indexed by issuer
+ */
+const isEqualPerIssuer = (strategies, leftQuantities, rightQuantities) =>
+  leftQuantities
+    .map((leftQ, i) => strategies[i].equals(leftQ, rightQuantities[i]))
+    .reduce(allTrue);
+
+/**
+ * `areRightConserved` checks that the total quantity per issuer stays
```
Chris-Hibbert: typo: areRightSConserved

core/zoe/isOfferSafe.js
```diff
@@ -0,0 +1,86 @@
+import { allTrue } from './utils';
+import { insist } from '../../util/insist';
+
+/**
+ * @param  {array} assaysPerIssuer - an array of assays ordered in the
+ * same order as the corresponding issuers
+ * @param  {array} rulesPerIssuer - an array of "rules" ordered in the
+ * same order as the corresponding issuers. Rules are a player's
+ * understanding of the contract that they are entering when they make
+ * an offer. Rules are structured in the form { rule: '', amount}
```
Chris-Hibbert: put something in the quotes, or replace it with a descriptive name like ruleDescriptionString.

core/zoe/isOfferSafe.js
```diff
@@ -0,0 +1,86 @@
...
+ * @param  {array} amountsPerIssuer - an array of amounts ordered in
+ * the same order as the corresponding issuers. This array of amounts
+ * is the reallocation to be given to a player.
+ *
+ * `isOfferSafeForPlayer` checks offer-safety for a single player.
```
Chris-Hibbert: This description should be above the list of parameters.

core/zoe/isOfferSafe.js
```diff
@@ -0,0 +1,86 @@
...
+    .reduce(allTrue, true);
+
+  // If we are not refunding the player, are their allocated amounts
+  // greater than or equal to what they said they wanted at the beginning?
+  const winningsOk = rulesPerIssuer
+    .map((rule, i) => {
+      // If the rule was 'wantExactly' or 'wantAtLeast', we should make
+      // sure that the user gets what they wanted. If it is not,
+      // anything we give back is fine.
+      if (rule.rule === 'wantExactly' || rule.rule === 'wantAtLeast') {
```
Chris-Hibbert: Do we ever treat wantExactly as different from wantAtLeast? Do we need both?

core/zoe/isOfferSafe.js
```diff
@@ -0,0 +1,86 @@
...
+    .reduce(allTrue, true);
+
+  return refundOk || winningsOk;
+}
+/**
+ * @param  {array} assays - An array of assays in the same order as
+ * their corresponding issuers
+ * @param  {matrix} offerMatrix - an array of arrays. Each of the
+ * element arrays is the offer (array of rules) that a single player
+ * made. In the same order as the corresponding issuers.
```
Chris-Hibbert: No sentence break. (and line 75)

test/unitTests/core/zoe/test-isOfferSafe.js
```diff
@@ -0,0 +1,236 @@
...
+    t.end();
+  }
+});
+
+// The user gets less than what they wanted - wantAtLeast
+test('isOfferSafeForPlayer - less than wantExactly', t => {
+  try {
+    const { assays } = setup();
+    const rules = [
+      { rule: 'haveExactly', amount: assays[0].make(2) },
```
Chris-Hibbert: There are several tests where all the amounts are in violation. How about at least one where only one amount is?

core/zoe/isOfferSafe.js
```diff
@@ -0,0 +1,86 @@
...
+    .map((rule, i) => {
+      // If the rule was 'haveExactly', we should make sure that the
+      // user gets it back in a refund. If it is not 'haveExactly' anything
+      // we give back is fine.
+      if (rule.rule === 'haveExactly') {
+        return isAmountOfferSafe(assaysPerIssuer[i], amountsPerIssuer[i], rule);
+      }
+      return true;
+    })
+    .reduce(allTrue, true);
```
Chris-Hibbert: ```.reduce()``` only seems to need one arg according to test-utils.  Same issue on line 63.

test/unitTests/core/zoe/test-utils.js
```diff
@@ -0,0 +1,189 @@
...
+    ];
+    t.notOk(offerEqual(assays, offer1, offer2));
+  } catch (e) {
+    t.assert(false, e);
+  } finally {
+    t.end();
+  }
+});
+
+test('offerEqual - returns false bc different rule', t => {
```
Chris-Hibbert: I'd like to see a test for WantExactly vs WantAtLeast. :-)

core/zoe/areRightsConserved.js
katelynsills: Good catch!

core/zoe/isOfferSafe.js
katelynsills: reduce takes an optional initial value, but I think you're right that `true` isn't needed here since we'll already get a good initial value from the first item.

katelynsills: Removed the initial value

core/zoe/isOfferSafe.js
katelynsills: Hmm, this is a really good question. I need to think on this. Since the "rule" is representing the user's contractual understanding, I think we need both so the user can express their exact intention. But that could mean that we should be rejecting getting more than what you wanted if the rule is 'wantExactly'. Hmm...

test/unitTests/core/zoe/test-utils.js
katelynsills: Added!

test/unitTests/core/zoe/test-isOfferSafe.js
katelynsills: Yup, added

core/zoe/isOfferSafe.js
katelynsills: Ok, now `isOfferSafeForPlayer` returns false if we are refunding more than `haveExactly` and if we are awarding winnings more than `wantExactly`. I added tests for those cases and cleaned up the tests a bit more

katelynsills:
Whoops, Mark reminded me that it will likely be `haveAtMost` rather than `haveAtLeast`. I'll need to make that change

katelynsills:
> Whoops, Mark reminded me that it will likely be `haveAtMost` rather than `haveAtLeast`. I'll need to make that change

Done

core/zoe/utils.js
```diff
@@ -0,0 +1,32 @@
+// used to reduce boolean arrays
+const bothTrue = (prev, curr) => prev && curr;
+const anyTrue = (prev, curr) => prev || curr;
+
+// https://stackoverflow.com/questions/17428587/transposing-a-2d-array-in-javascript/41772644#41772644
+const transpose = matrix =>
```
erights: Do you actually need both representations, or could you just keep it in the transposed form?

core/zoe/utils.js
```diff
@@ -0,0 +1,32 @@
...
+  matrix.reduce(
+    (acc, row) => row.map((_, i) => [...(acc[i] || []), row[i]]),
+    [],
+  );
+
+const ruleEqual = (leftRule, rightRule) => leftRule.rule === rightRule.rule;
```
erights: I would be very surprised if `===` is what you want here, rather than `sameStructure`

core/zoe/utils.js
```diff
@@ -0,0 +1,32 @@
...
+    return false;
+  }
+  return leftOffer
+    .map((leftRule, i) => {
+      return (
+        ruleEqual(leftRule, rightOffer[i]) &&
+        amountEqual(assays[i], leftRule, rightOffer[i])
+      );
+    })
+    .reduce(bothTrue);
```
erights: If both lengths are 0, they're equal.

core/zoe/utils.js
erights: No, this is actually correct. From the names, I thought you were comparing rules. But you're actually comparing strings representing the kind of rule. `===` is good. Resolving.

core/zoe/isOfferSafe.js
```diff
@@ -0,0 +1,105 @@
...
+      // If the rule is something else, anything we give back is fine.
+      if (rule.rule === 'wantExactly') {
+        return assaysPerIssuer[i].equals(amountsPerIssuer[i], rule.amount);
+      }
+      if (rule.rule === 'wantAtLeast') {
+        return assaysPerIssuer[i].includes(amountsPerIssuer[i], rule.amount);
+      }
+      return true;
+    })
+    .reduce(bothTrue);
```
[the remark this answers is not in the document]
katelynsills: Done!

core/zoe/isOfferSafe.js
```diff
@@ -0,0 +1,105 @@
...
+      // we give back is fine.
+      if (rule.rule === 'haveExactly') {
+        return assaysPerIssuer[i].equals(amountsPerIssuer[i], rule.amount);
+      }
+      if (rule.rule === 'haveAtMost') {
+        return assaysPerIssuer[i].includes(amountsPerIssuer[i], rule.amount);
+      }
+      return true;
+    })
+    .reduce(bothTrue);
```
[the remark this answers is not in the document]
katelynsills: Good point! I had had `reduce(bothTrue, true)` in some places and somehow they got taken out. I've added tests for the empty cases.

core/zoe/isOfferSafe.js
```diff
@@ -0,0 +1,105 @@
...
+  amountsPerIssuer,
+) {
+  insist(
+    assaysPerIssuer.length === rulesPerIssuer.length &&
+      assaysPerIssuer.length === amountsPerIssuer.length,
+  )`assays, rules, and amounts must be arrays of the same length`;
+
+  const allowedRules = [
+    'haveExactly',
+    'haveAtMost',
```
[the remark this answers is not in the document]
katelynsills: Darn, this got changed to `offerExactly` and `offerAtMost` in the next PR up the stack and didn't get pulled down to this one. I'll rectify that.

core/zoe/utils.js
```diff
@@ -0,0 +1,32 @@
+// used to reduce boolean arrays
+const bothTrue = (prev, curr) => prev && curr;
+const anyTrue = (prev, curr) => prev || curr;
```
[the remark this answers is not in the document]
katelynsills: Good point.

core/zoe/isOfferSafe.js
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * `isOfferSafeForPlayer` checks offer-safety for a single player.
+ *
+ * Note: This implementation checks whether we refund for all rules or
+ * return winnings for all rules. It does not allow some refunds and
+ * some winnings, which is what would happen if you checked the rules
+ * independently.
+ *
+ * @param  {array} assaysPerIssuer - an array of assays ordered in the
```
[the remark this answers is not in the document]
katelynsills: Very good point. I was originally thinking that offer safety included an issuer check (which operating on quantities using strategies doesn't give you) but I think we can safely assume that that is done elsewhere in Zoe, when the payments are made using the array of issuers. 

I was able to change to quantities/strategies, so that change is in, at least for this PR!

core/zoe/isOfferSafe.js
```diff
@@ -0,0 +1,105 @@
...
+
+  return refundOk || winningsOk;
+}
+/**
+ * @param  {array} assays - An array of assays in the same order as
+ * their corresponding issuers
+ * @param  {matrix} offerMatrix - an array of arrays. Each of the
+ * element arrays is the offer (array of rules) that a single player
+ * made, in the same order as the corresponding issuers.
+ * @param  {matrix} amountMatrix - an array of arrays. Each of the
```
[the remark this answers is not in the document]
katelynsills: Amounts and quantities matrices, you mean? I would agree. We still do need to transpose though.

core/zoe/utils.js
katelynsills: We only need the transposed version to do the conservation of rights check, because that's summing by issuer column, whereas most everything else is concerned with a player perspective (the rows). Since we don't actually store both, I think this is ok. Did you think this would give bad performance?

core/zoe/isOfferSafe.js
```diff
@@ -0,0 +1,105 @@
...
+    'wantExactly',
+    'wantAtLeast',
+  ];
+
+  // If we are refunding the player, are their allocated amounts
+  // greater than or equal to what they said they had at the beginning?
+  const refundOk = rulesPerIssuer
+    .map((rule, i) => {
+      insist(
+        allowedRules.includes(rule.rule),
```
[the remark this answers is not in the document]
katelynsills: That's a good idea. Should we support `null` and `undefined` as `offerDescElem`s or only `null`? In general, are there benefits to preferring one over the other? I'm allowing both for now.

katelynsills:
I'm closing this PR in favor of a new set of Zoe PRs in which this one (#123), #125, and #136 are squashed and then pulled apart again.

PR #125, opened 2019-09-19
Add Zoe and Basic Contracts

katelynsills:
This PR adds Zoe plus three very basic contracts:

1. Automatic Refund: a trivial contract that gives the user back what they put in.
2. Simple Offer Swap: a swap contract in which there is only one method in the governing contract API: `makeOffer`. The governing contract instance must be passed around to the various parties for them to make their offer. There are no invites to make an offer.
3. Simple Offer Second-price Auction: a second-price auction contract that uses the same "middle layer" as Simple Offer Swap, but makes changes to the logic of when an offer is valid and how the quantities should be reallocated such that it's a second price auction rather than a swap.

AutomaticRefund can be used to explain the basics of Zoe and for testing, but isn't actually useful. 

The Simple Offer Swap and the Simple Offer Second-price Auction are useful and show that we can build reusable code on top of Zoe such that only the "business logic" of the particular institution changes. 

There's [documentation on the swap and second-price auction](https://github.com/Agoric/ERTP/tree/zoe-basic-contracts/core/zoe/docs), but more documentation is needed.

Still to be done:
* Porting to a SwingSet vat environment and writing tests for that.
* Writing up documentation on the why of Zoe, and the Zoe APIs for governing contracts and end users.


core/zoe/contracts/simpleOffer/simpleOffer.js
```diff
@@ -0,0 +1,54 @@
+import harden from '@agoric/harden';
+
+import makePromise from '../../../../util/makePromise';
+
+const makeSimpleOfferMaker = srcs => zoeInstance => {
```
Chris-Hibbert: The layering here needs more documentation. If SWAPS.md is intended to document the use of this file, then there should be a link here.

Please describe the expectations of isValidOffer(), reallocate(), and canReallocate(). 

If this hides zoe from the contracts above it, then it should say more about how they interact.

Is the use of simpleOffer apparent to the client of a second price auction? What do they see?

test/unitTests/core/zoe/contracts/test-simpleOfferSwap.js
```diff
@@ -0,0 +1,139 @@
+import { test } from 'tape-promise/tape';
+import harden from '@agoric/harden';
+
+import { makeZoe } from '../../../../../core/zoe/zoe/zoe';
+import { makeSimpleOfferMaker } from '../../../../../core/zoe/contracts/simpleOffer/simpleOffer';
+import { swapSrcs } from '../../../../../core/zoe/contracts/simpleOffer/srcs/swapSrcs';
+import { setup } from '../setupBasicMints';
+
+test('zoe.makeInstance with simpleOffer with swapSrcs', async t => {
```
Chris-Hibbert: There should be a test that re-uses a zoe instance.

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,163 @@
...
+}
+
+/**
+ * This is returned by a call to `escrow` on the zoeInstance userFacet.
+ * An escrowReceipt is an ERTP payment which acts as proof of escrow
+ * and is normally passed in to a governingContract method. A
+ * claimPayoff ERTP payment is what can be unwrapped to claim the
+ * payoff of involvement with a contract. 
+ */
+struct UserFacetAndGoverningContract ( ) {
```
Chris-Hibbert: wrong name, I think you mean EscrowReceiptAndClaimPayoff.

When we teach the chainmail parser to do something with its parse tree, we'll catch this redundant declaration.

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * Zoe is a long-lived and well-trusted contract that enforces
+ * offer-safety for the contracts that use it. Zoe has a single
+ * `seatIssuer` and a single `escrowReceiptIssuer` for the entirety of its
+ * lifetime. By having a reference to Zoe, a user can get the `seatIssuer`
+ * or `escrowReceiptIssuer` and thus validate any `seat` or `escrowReceipt`
+ * they receive from someone else. 
+ */
+
+
```
Chris-Hibbert: remove blank lines

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,163 @@
...
+interface Zoe ( ) {
+  /** 
+   * Zoe is long-lived, but we want to create individual zoeInstances
+   * that are instantiated with a particular governingContract and
+   * particular issuers. For example, a zoeInstance may be instantiated
+   * with a swap contract and an array of the issuers for moola and
+   * simoleans. We get back an object that has the user facet for the
+   * zoeInstance and the governing contract facet. `makeContract` is a
+   * function that takes a zoeInstance
```
Chris-Hibbert: Add a period at the end so it doesn't look like we were thinking about saying more.

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,163 @@
...
+   * `payments` is an array of the actual payments to be escrowed,
+   * following the rules in the offerDesc. If the offerDesc rule is
+   * 'haveExactly' or 'haveAtMost', then a payment matching the amount is
+   * expected. If the offerDesc rule is 'wantAtLeast' or 'wantExactly'
+   * then the payment will be ignored and should be undefined.
+   */
+  escrow (offerDesc :List, payments :List) -> (escrowReceiptAndClaimPayoff);
+
+  /**
+   * Returns the array of issuers that the zoeInstance was instantiated
```
Chris-Hibbert: the zoeInstance was instantiated with seems repetitive to me. How about 'Returns the zoeInstance's array of issuers'?

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,163 @@
...
+ * Zoe is a long-lived and well-trusted contract that enforces
+ * offer-safety for the contracts that use it. Zoe has a single
+ * `seatIssuer` and a single `escrowReceiptIssuer` for the entirety of its
+ * lifetime. By having a reference to Zoe, a user can get the `seatIssuer`
+ * or `escrowReceiptIssuer` and thus validate any `seat` or `escrowReceipt`
+ * they receive from someone else. 
+ */
+
+
+interface Zoe ( ) {
```
Chris-Hibbert: When the interface doesn't have type parameters, leave off the empty parens.

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,163 @@
...
+interface ZoeGoverningContractFacet () {
+
+  /** 
+   * Instruct Zoe to try reallocating for the given offerIds.
+   * Reallocation is a matrix (array of arrays) where the rows are the
+   * quantities to be paid to the player who made the offer at the same
+   * index in the offerIds array. The reallocation will only happen if
+   * offer safety and conservation of rights are true, as enforced by Zoe.
+   */
+  reallocate (offerIds :List, reallocation :List);
```
Chris-Hibbert: reallocation :List(List(Quantity))

IIUC

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,163 @@
...
+   * expected. If the offerDesc rule is 'wantAtLeast' or 'wantExactly'
+   * then the payment will be ignored and should be undefined.
+   */
+  escrow (offerDesc :List, payments :List) -> (escrowReceiptAndClaimPayoff);
+
+  /**
+   * Returns the array of issuers that the zoeInstance was instantiated
+   * with. This is helpful in the case that the user has forgotten the order.
+   */
+  getIssuers() -> (List);
```
Chris-Hibbert: List(Issuer)

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,163 @@
...
+   * property. The objects in the `offerDesc` must be in the same order
+   * as the issuers array. That is, the amount in index 0 of `offerDesc`
+   * should be an amount for the issuer in index 0 of the issuers array.
+   * `payments` is an array of the actual payments to be escrowed,
+   * following the rules in the offerDesc. If the offerDesc rule is
+   * 'haveExactly' or 'haveAtMost', then a payment matching the amount is
+   * expected. If the offerDesc rule is 'wantAtLeast' or 'wantExactly'
+   * then the payment will be ignored and should be undefined.
+   */
+  escrow (offerDesc :List, payments :List) -> (escrowReceiptAndClaimPayoff);
```
Chris-Hibbert: Can you write a declaration for rules and make this be

```
escrow(offerDesc :List(Rule), payments :List(Payment)) -> (EscrowReceiptAndClaimPayoff)
```

core/zoe/docs/SecondPriceAuction.md
```diff
@@ -0,0 +1,81 @@
...
+  escrowReceipt: allegedAliceEscrowReceipt,
+  claimWinnings: aliceClaimWinnings,
+} = await zoeInstance.escrow(aliceOfferDesc, alicePayments);
+```
+
+Note that in this implementation, the item that will be auctioned is
+described at index 0, and Alice's minimum bid amount is at index 1 in
+the offer description. 
+
+Now Alice can spread her auction instance far and wide and see if
```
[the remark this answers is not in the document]
katelynsills: For this version that uses the simpleOffer interface, she shares the auction instance. You could have another version that uses invites.

core/zoe/zoe/zoeUtils.js
```diff
@@ -0,0 +1,115 @@
...
+};
+
+const depositAll = async (purses, strategies, offerDesc, offerPayments) => {
+  const quantitiesArrayPromises = purses.map(async (purse, i) => {
+    // if the user's contractual understanding includes
+    // "haveExactly" or "haveAtMost", make sure that they have supplied a
+    // payment with that exact balance
+    if (
+      offerDesc[i].rule === 'haveExactly' ||
+      offerDesc[i].rule === 'haveAtMost'
```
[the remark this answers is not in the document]
katelynsills: I like that

test/unitTests/core/zoe/contracts/test-simpleOfferSwap.js
katelynsills: What kind of re-use?

core/zoe/contracts/simpleOffer/srcs/secondPriceSrcs.js
```diff
@@ -0,0 +1,107 @@
...
+
+const reallocate = (
+  strategies,
+  offerIds,
+  offerIdsToOfferDescs,
+  getQuantitiesFor,
+) => {
+  const bids = offerIds
+    .map(offerId => offerIdsToOfferDescs.get(offerId)[1].amount.quantity)
+    .slice(1);
```
Chris-Hibbert: comment that this is to remove the person who created the auction. (IIUC)

Shouldn't slice() be called on offerIds rather than the result of the map() call?

core/zoe/contracts/simpleOffer/srcs/secondPriceSrcs.js
```diff
@@ -0,0 +1,107 @@
...
+  const { winner, price } = findWinnerAndPrice(
+    strategies[1],
+    offerIds.slice(1),
+    bids,
+  );
+
+  const creator = offerIds[0];
+  const creatorQuantities = getQuantitiesFor(harden([creator]))[0];
```
Chris-Hibbert: All (nearly all? I can't tell) of these zeroes and ones should be replaced with manifest constants indicating that 0 is the itemIndex, and 1 is the priceIndex.  Above, there are cases where 0 is the sellerIndex (and some 1s are really `sellerIndex + 1`. I'm not sure what other 0s and 1s might be in use, but it'll be easier to see once these have names.

core/zoe/contracts/simpleOffer/srcs/secondPriceSrcs.js
```diff
@@ -0,0 +1,107 @@
...
+
+  // The winner gets the assets put up for auction.
+  // eslint-disable-next-line prefer-destructuring
+  winnerQuantities[0] = creatorQuantities[0];
+  creatorQuantities[0] = strategies[1].empty();
+
+  // The person who created the auction gets the price paid.
+  creatorQuantities[1] = price;
+
+  // The winner gets to keep the difference between the their bid and
```
Chris-Hibbert: typo

test/unitTests/core/zoe/contracts/test-simpleOfferAuction.js
```diff
@@ -0,0 +1,252 @@
...
+
+    // Setup Dave
+    const daveMoolaPurse = mints[0].mint(issuers[0].makeAmount(0));
+    const daveMoolaPayment = daveMoolaPurse.withdrawAll();
+    const daveSimoleanPurse = mints[1].mint(issuers[1].makeAmount(5));
+    const daveSimoleanPayment = daveSimoleanPurse.withdrawAll();
+
+    // 1: Alice creates a secondPriceAuction instance
+    const secondPriceSrcs = makeSecondPriceSrcs(3);
+    const makeSecondPriceAuction = makeSimpleOfferMaker(secondPriceSrcs);
```
Chris-Hibbert: I don't understand why the packaging is done this way. Alice puts the simpleOfferMaker and secondPrice code together.  Would it make sense to have an auctionMaker that did that packaging, so Alice doesn't have to?

core/zoe/contracts/simpleOffer/srcs/secondPriceSrcs.js
```diff
@@ -0,0 +1,107 @@
...
+  const isInitialOffer = offerIds.length === 0;
+  return (
+    (isInitialOffer && isValidInitialOfferDesc(offerMadeDesc)) ||
+    (!isInitialOffer &&
+      isValidBid(assays, offerIdsToOfferDescs.get(offerIds[0]), offerMadeDesc))
+  );
+};
+
+// Iterate over all of the bids keeping the highest and second highest bid.
+const findWinnerAndPrice = (strategy, offerIds, bids) => {
```
Chris-Hibbert: This code assumes that the `initialOffer` isn't included in `offerIds`. Would it make sense to rename it `bidIds`?

test/unitTests/core/zoe/contracts/test-simpleOfferSwap.js
Chris-Hibbert: The main use of a zoe is to make an instance of a contract with `makeInstance`. Zoe is supposed to be long-lived and ready to support multiple instances of multiple contracts. I'd like to see a test that different contract instances don't interfere with each other.

Chris-Hibbert:
Two more things I thought of overnight: 

* I'd like to see a mutli-vat test. It's important to show that Zoe and the framework will work when the (renamed) issuers are in separate vats. I'm hopeful that the ability to import the extentOps will reduce the pain of that over what I had to work with for my auction.

* How does Bob validate the code that Alice is running for Zoe, the simpleSwap, and the contract?

katelynsills:
> I'd like to see a mutli-vat test. It's important to show that Zoe and the framework will work when the (renamed) issuers are in separate vats. I'm hopeful that the ability to import the extentOps will reduce the pain of that over what I had to work with for my auction.

I would too :) I'm waiting on E being available outside of SwingSet before I write Swingset tests so that my unit tests don't all break when I start adding E() to Zoe. Before we can say that Zoe is usable, we must have SwingSet tests, but I think this could be pushed into a PR that is different than this stack of PRs. What do you think?

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,163 @@
...
+   * the zoeInstance.
+   */
+  escrowOffer (offerDesc :List, offerPayments :List) -> (offerId);
+
+  /** 
+   * A helper function provided by Zoe to allow the governing contract
+   * to easily burn and validate an escrowReceipt received from the
+   * user. 
+   */
+  burnEscrowReceipt (escrowReceipt :Payment) -> (Quantity);
```
Chris-Hibbert: I see several places  (here and [elsewhere](https://github.com/Agoric/ERTP/pull/136/files)) where the result of `burnEscrowReceipt()` is destructured as `{ id, offerMade }`, and then `id` is used as an index into offers. Why does this work?

Chris-Hibbert:
on testing with vats: 

My experience with the second price auction I wrote (currently blocked by the Presence bug) leads me to believe that the code will get much more convoluted in order to deal with sending Issuers and AssetHolders back and forth among multiple vats. 

Zoe has an advantage in that the contracts don't deal with AssetHolders directly, but Zoe does and she has to convert to assetDescs and extents. It may be that the importable descOps and extentOps will make that tractable, but we don't have any actual experience with them. I expect you'll find ways to simplify all these interactions, but I expect the consequences to be significant, and they might even lead to architectural changes to make the code readable.

Yeah, it would be nice to get through this stack of PRs and have something stable to start that refactoring on.

test/unitTests/core/zoe/contracts/test-simpleOfferSwap.js
katelynsills: Ah, that makes sense. Sure, I can do that

core/zoe/docs/SWAPS.md
```diff
@@ -0,0 +1,78 @@
+# Swaps
+
+If I want to trade one kind of asset for another kind, I could send
+you the asset and ask you to send me the other kind back. But, you
+could choose to behave opportunistically: receive my asset and give
+nothing back. To solve this problem, swap contracts allow users to
+securely trade one kind of eright for another kind, leveraging Zoe
+for escrow and offer-safety. At no time does any user have the
+opportunity to behave opportunistically.
```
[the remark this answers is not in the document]
katelynsills: Hmm, I've been taught that repeating the same word ("opportunistically") can often be desirable in explanatory writing because using a slightly different word makes the user wonder if there's a substantial difference between "cheating" and "acting opportunistically." I like the change to `ability` though. I'll make that change. Would it be possible to separate copy-editing from factual errors and do a later run through all the documentation for things that aren't factual errors but may be improved by rewording?

core/zoe/contracts/simpleOffer/srcs/secondPriceSrcs.js
```diff
@@ -0,0 +1,107 @@
+import harden from '@agoric/harden';
+
+import { makeHasOkRules } from '../../../contractUtils';
+
+const hasOkRulesInitialOffer = makeHasOkRules([
+  ['haveExactly', 'wantAtLeast'], // The initial offer
+]);
+
+const hasOkRulesBids = makeHasOkRules([
+  ['wantExactly', 'haveAtMost'], // Subsequent bids
```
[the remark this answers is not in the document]
katelynsills: Yeah, ok, I think you're right. I liked the easy dichotomy of `have` vs `want` but you're right that saying `haveAtMost` doesn't make sense. You either have it or you don't.

core/zoe/contractUtils.js
```diff
@@ -0,0 +1,76 @@
+// These utilities are likely to be helpful to developers writing
+// governing contracts.
+
+// used to reduce boolean arrays
+const allTrue = (prev, curr) => prev && curr;
```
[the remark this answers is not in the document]
katelynsills: Yeah, that makes sense. I'll change it

core/zoe/zoe/state.js
```diff
@@ -0,0 +1,47 @@
...
+  const quantities = new WeakMap();
+  const offers = new WeakMap();
+  const results = new WeakMap();
+  const purses = issuers.map(issuer => issuer.makeEmptyPurse());
+
+  const readOnlyState = harden({
+    issuers,
+    assays: issuers.map(issuer => issuer.getAssay()),
+    strategies: issuers.map(issuer => issuer.getStrategy()),
+    getIssuers: () => issuers,
```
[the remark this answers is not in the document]
katelynsills: Good point

core/zoe/zoe/zoeUtils.js
```diff
@@ -0,0 +1,115 @@
...
+  const claimWinningsPurseP = seatMint.mint(claimWinningsQuantity);
+  const seat = harden({
+    getWinnings: () => result.p,
+  });
+  addUseObj(claimWinningsQuantity.id, seat);
+  const claimWinningsPaymentP = claimWinningsPurseP.withdrawAll();
+  return claimWinningsPaymentP;
+};
+
+const depositAll = async (purses, strategies, offerDesc, offerPayments) => {
```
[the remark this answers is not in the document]
katelynsills: I like that

core/zoe/zoe.chainmail
katelynsills: Yup, good catch

core/zoe/zoe.chainmail
katelynsills: Yup!

core/zoe/zoe.chainmail
katelynsills: Yup!

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,163 @@
...
+
+  /** 
+   * Eject the offer, taking the current allocation for that offer and
+   * creating payments to be returned to the user. The payments can be
+   * claimed by the user by taking the `claimPayment` ERTP payment
+   * received after escrowing, calling `unwrap` and calling the `claimPayoff`
+   * method on the resulting use object. No offer safety checks are
+   * done here because any previous reallocation performed those checks.
+   */
+  eject (offerIds :List);
```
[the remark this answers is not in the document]
katelynsills: Hmm, good point. I think I like `complete` the best out of that list. I wanted to emphasize two things: the player is given their payments (either refunds or winnings) and the player is kicked out and can't participate further without a new offer.

core/zoe/zoe/zoe.js
```diff
@@ -0,0 +1,203 @@
...
+      // couple of things. It can propose a reallocation of
+      // quantities, eject an offer, and interestingly, can create a
+      // new offer itself for recordkeeping and other various
+      // purposes.
+
+      const governingContractFacet = harden({
+        /**
+         * The governing contract can propose a reallocation of
+         * quantities per player, which will only succeed if the
+         * reallocation 1) conserves rights, and 2) is offer safe for
```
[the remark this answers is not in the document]
katelynsills: Good point

core/zoe/contracts/simpleOffer/srcs/secondPriceSrcs.js
```diff
@@ -0,0 +1,107 @@
...
+  winnerQuantities[1] = strategies[1].without(winnerQuantities[1], price);
+
+  // Everyone else gets a refund so their quantities remain the same.
+  return harden({
+    reallocOfferIds: [creator, winner],
+    reallocQuantities: [creatorQuantities, winnerQuantities],
+  });
+};
+
+const makeSecondPriceSrcs = numBids =>
```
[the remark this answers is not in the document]
katelynsills: They're plain JavaScript modules now, but that won't work in a SwingSet environment, so the name `srcs` is meant to indicate that they're pure functions that should be passed safely across vats, through whatever mechanism we decide on. It seems like we're still figuring out our long-term solution for that, so I left the name as is. The next step would be to use the import manager. My understanding of the srcs/evaluate pattern in the contract Host is that that was our method before we came up with the import manager. Is there any reason why we would keep that pattern around?

test/unitTests/core/zoe/contracts/test-automaticRefund.js
```diff
@@ -0,0 +1,160 @@
...
+    // 5: Imagine that Bob also has access to the zoeInstance and the
+    // automaticRefund
+
+    // 6: Bob also wants to get an automaticRefund (why? we don't
+    // know) so he escrows his offer and his offer payments.
+
+    const bobOfferDesc = harden([
+      {
+        rule: 'wantExactly',
+        amount: issuers[0].makeAmount(3),
```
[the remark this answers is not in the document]
katelynsills: Good point

core/zoe/contracts/simpleOffer/srcs/secondPriceSrcs.js
katelynsills: Yeah, there are multiple different reasons for indexing into an array with zero. You're right that it's confusing. Let me see what I can do

katelynsills: Ok I think I pushed up those changes. Let me know what you think

core/zoe/contracts/simpleOffer/srcs/secondPriceSrcs.js
katelynsills: Yup

core/zoe/contracts/simpleOffer/srcs/secondPriceSrcs.js
katelynsills: Yes

core/zoe/zoe/zoeUtils.js
```diff
@@ -0,0 +1,115 @@
...
+
+const makePayments = (purses, amountsMatrix) =>
+  amountsMatrix.map(row =>
+    row.map((amount, i) => {
+      const payment = purses[i].withdraw(amount, 'payout');
+      return payment;
+    }),
+  );
+
+const escrowEmptyOffer = (adminState, assays, strategies) => {
```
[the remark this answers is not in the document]
katelynsills: I took another look and it was easiest to have escrowEmptyOffer prepare the offerDesc and payments and then pass that to escrowOffer. The payments will be ignored because all of the rules for the empty offer are `wantAtLeast`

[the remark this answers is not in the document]
katelynsills: There's a little bit of unnecessary computation, but we get the benefit of reusing the same code path, which I think is better.

core/zoe/zoe.chainmail
katelynsills: ah! Because the quantity is an object with the properties { id, offerMade }. I had to double check that just now because it didn't make sense to me either.

core/zoe/zoe.chainmail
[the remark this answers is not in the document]
katelynsills: Ok, I changed `eject` to `complete`. I think it works well!

core/zoe/zoe/zoeUtils.js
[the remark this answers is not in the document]
Chris-Hibbert: looks good.

test/unitTests/core/zoe/contracts/test-simpleOfferSwap.js
katelynsills: Ok, I added a test to the automaticRefund tests that creates three automatic refunds from the same Zoe and sends one offer to each. I changed automaticRefund to keep a count of the offers it receives such that the test can check that the offer count is 1 for each instance. 

It's not rigorous but should do for now I think. Dean had an idea for how Zoe should work that would move to eliminating the "zoeInstance" and just having the Zoe and the governing contract, so it probably doesn't make sense to put all our effort into proving things now.

test/unitTests/core/zoe/contracts/test-simpleOfferAuction.js
katelynsills: Yeah, I agree that it's rough. I think the "packaging" issue is related to how Alice can reuse highly-vetted governing contracts from libraries. That is, if Alice just looks up a name, this will turn out pretty different. Is it ok if we punt this until we can have a discussion on how to share reusable code, use it easily, and convince other people that you used a particular library?

Chris-Hibbert: certainly.

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,165 @@
...
+ * offer-safety for the contracts that use it. Zoe has a single
+ * `seatIssuer` and a single `escrowReceiptIssuer` for the entirety of its
+ * lifetime. By having a reference to Zoe, a user can get the `seatIssuer`
+ * or `escrowReceiptIssuer` and thus validate any `seat` or `escrowReceipt`
+ * they receive from someone else. 
+ */
+
+interface Zoe {
+  /** 
+   * Zoe is long-lived, but we want to create individual zoeInstances
```
erights: The word "instance" here is confusing. The long-lived credible thing that Alice and Bob both know before they know each other is indeed an instance of this interface. That instance has a method, currently called `makeInstance`, that is making something specific to a given contract that Alice and Bob will participate it. Let's call this a Zoe Pool. In the previous scenario, Alice creates a Zoe pool and communicates something specific to this pool to Bob for Bob to participate. Bob need to validate that something against the long lived Zoe instance, i.e., the object that implements the `Zoe` interface.

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,165 @@
...
+   * To escrow, the user must provide an `offerDesc` which is a list of
+   * objects with a rule property (such as 'offerExactly') and an amount
+   * property. The objects in the `offerDesc` must be in the same order
+   * as the issuers array. That is, the amount in index 0 of `offerDesc`
+   * should be an amount for the issuer in index 0 of the issuers array.
+   * `payments` is an array of the actual payments to be escrowed,
+   * following the rules in the offerDesc. If the offerDesc rule is
+   * 'offerExactly' or 'offerAtMost', then a payment matching the amount is
+   * expected. If the offerDesc rule is 'wantAtLeast' or 'wantExactly'
+   * then the payment will be ignored and should be undefined.
```
erights: > should be undefined

Good. The example code should use undefined in that position.

In fact, perhaps if it is anything other than undefined it is an error.

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,165 @@
...
+   * index in the offerIds array. The reallocation will only happen if
+   * 'offer safety' and conservation of rights are true, as enforced by Zoe.
+   */
+  reallocate (offerIds :List, reallocation :List(List(Quantity)));
+
+  /** 
+   * Eject the offer, taking the current allocation for that offer and
+   * creating payments to be returned to the user. The payments can be
+   * claimed by the user by taking the `claimPayment` ERTP payment
+   * received after escrowing, calling `unwrap` and calling the `claimPayoff`
```
erights: If `claimPayoff` a method or a property? The chainmail puts it in a pass-by-copy struct and therefore declares `claimPayoff` as a property.

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,165 @@
...
+   * Escrow an offer created by the governing contract. Autoswap uses
+   * this to mint liquidity tokens and add them to the rights managed by
+   * the zoeInstance.
+   */
+  escrowOffer (offerDesc :List, offerPayments :List) -> (offerId);
+
+  /** Burn and validate an escrowReceipt received from the user. */
+  burnEscrowReceipt (escrowReceipt :Payment) -> (Quantity);
+
+  ////// The methods below are pure and have no side-effects. ////////
```
erights: The methods indeed have no side effects. But, for example, `getIssuers` is not pure because it gives you issuers which are not pure.

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,165 @@
...
+
+  /**
+   * Create an array of empty quantities per issuer. Note that if the 
+   * mint is not a basic fungible mint, this may be something other than 0. 
+   */
+  makeEmptyQuantities ( ) -> (List(Quantity));
```
erights: pure

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,165 @@
...
+
+  /** Get the array of issuers */
+  getIssuers ( ) -> (List(Issuer));
```
erights: not pure

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,165 @@
...
+  
+  /** Get the array of assays */
+  getAssays ( ) -> (List(Assay));
```
erights: pure

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,165 @@
...
+
+  /** Get the array of strategies, the logic from the assays */
+  getStrategies ( ) -> (List(Strategy));
```
erights: pure

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,165 @@
...
+
+  /** 
+   * Pass in an array of offerIds and get a matrix (array of arrays)
+   * containing the quantities, in the same order as the offerIds array.
+   */
+  getQuantitiesFor (offerIds :List) -> (List(List(Quantity)));
```
erights: pure

erights: Whenever coherent, you should expand `List` declarations to `List(SomethingMeaningful)`

core/zoe/zoe/state.js
```diff
@@ -0,0 +1,46 @@
+import harden from '@agoric/harden';
+
+const makeState = issuers => {
+  const quantities = new WeakMap();
+  const offers = new WeakMap();
+  const results = new WeakMap();
+  const purses = issuers.map(issuer => issuer.makeEmptyPurse());
```
erights: Be aware that you will have to do an additional asynchrony dance on initialization.

core/zoe/zoe/state.js
```diff
@@ -0,0 +1,46 @@
...
+      // has-side-effects
+      // eslint-disable-next-line array-callback-return
+      objIds.map(objId => {
+        quantities.delete(objId);
+        offers.delete(objId);
+        results.delete(objId);
+      });
+    },
+  });
+  return {
```
erights: Nice factoring!

core/zoe/zoe/zoe.js
```diff
@@ -0,0 +1,207 @@
...
+
+        burnEscrowReceipt: async escrowReceipt => {
+          const amount = await escrowReceiptIssuer.burnAll(escrowReceipt);
+          return amount.quantity;
+        },
+
+        // read-only, side-effect-free access below this line:
+        makeEmptyQuantities: () =>
+          makeEmptyQuantities(readOnlyState.getStrategies()),
+        getStrategies: readOnlyState.getStrategies,
```
erights: Possibilities:

...readOnlyState?

Treat readOnlyState as a trait and reuse the trait composition pattern?

provide a getReadOnlyState() method whose result is the readOnlyState object itself.

I think I prefer the last.

core/zoe/docs/SWAPS.md
```diff
@@ -0,0 +1,78 @@
...
+  {
+    rule: 'offerExactly',
+    amount: moolaIssuer.makeAmount(3),
+  },
+  {
+    rule: 'wantExactly',
+    amount: simoleanIssuer.makeAmount(7),
+  },
+]);
+const alicePayments = [aliceMoolaPayment, aliceSimoleanPayment];
```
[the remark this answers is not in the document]
katelynsills: Good point. Alice doesn't need to send in a simoleans payment object. She can send in `undefined` for that element in the array. If she passes in a payment, it will be ignored. I should remedy this in the tests and documents

core/zoe/zoe/state.js
katelynsills: yup! That is expected

core/zoe/contractUtils.js
```diff
@@ -0,0 +1,80 @@
...
+// https://stackoverflow.com/questions/17428587/transposing-a-2d-array-in-javascript/41772644#41772644
+const transpose = matrix =>
+  matrix.reduce(
+    (acc, row) => row.map((_, i) => [...(acc[i] || []), row[i]]),
+    [],
+  );
+
+/**
+ * @param  {matrix} matrix - array of arrays
+ * @param  {function} arrayF - the array of functions to apply
```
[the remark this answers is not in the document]
katelynsills: good call

core/zoe/contractUtils.js
```diff
@@ -0,0 +1,80 @@
+// These utilities are likely to be helpful to developers writing
+// governing contracts.
+
+// used to reduce boolean arrays
+const bothTrue = (prev, curr) => prev && curr;
+const eitherTrue = (prev, curr) => prev || curr;
```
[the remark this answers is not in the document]
katelynsills: good catch. I was able to get rid of `bothTrue` and `eitherTrue` entirely

core/zoe/contractUtils.js
```diff
@@ -0,0 +1,80 @@
...
+};
+
+// an array of empty quantities per strategy
+const makeEmptyQuantities = strategies =>
+  strategies.map(strategy => strategy.empty());
+
+// validRules is an array of arrays where each row is the rules of a valid offer:
+// e.g. validRules =
+//     [['offerExactly', 'wantExactly'], ['wantExactly', 'offerExactly']]
+const makeHasOkRules = validRules => offer =>
```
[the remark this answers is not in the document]
katelynsills: Sure, that sounds good. What does this mean? 

> Roles distinguished by invite rather than sniffing rules.

core/zoe/contracts/simpleOffer/srcs/swapSrcs.js
```diff
@@ -0,0 +1,63 @@
...
+      rule: firstOffer[1].rule,
+      amount: firstOffer[0].amount,
+    },
+    {
+      rule: firstOffer[0].rule,
+      amount: firstOffer[1].amount,
+    },
+  ]);
+
+const isValidFirstOfferDesc = newOfferDesc => hasOkRules(newOfferDesc);
```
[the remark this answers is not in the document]
katelynsills: It's almost that. the array of valid rules would be passed into `makeHasOkRules`, and the `newOfferDesc` is passed into the resulting `hasOkRules`. This change is in!

core/zoe/contracts/simpleOffer/srcs/secondPriceSrcs.js
```diff
@@ -0,0 +1,134 @@
...
+
+import { makeHasOkRules } from '../../../contractUtils';
+
+// The issuer array is ordered such that the item issuer is first
+// and the price issuer is second. All of the related arrays
+// (assays, strategies, etc.) also use this same ordering.
+const ITEM_INDEX = 0;
+const PRICE_INDEX = 1;
+
+// We expect the first offer to be the creator of the auction
```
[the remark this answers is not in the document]
katelynsills: I have a version of the swap with invites (not in a PR yet since I was trying to land uniswap first, but it's the one we went over on the big screen), but I was hoping to show a usable version without invites, so that `simpleOffer` is really as simple as possible. 

I think you're 100% right on not relying on the order of offer to indicate position. Let me see if I can improve that. It may be that the non-invite model just doesn't work, but since it is so much simpler I'd like to prove that to ourselves. 

One difference in how we may be thinking about it: I wasn't envisioning that an Arthur the auctioneer existed, since it's trivial for anyone who wants to sell something at auction to make the auction themselves. I'm not sure what role Arthur would play or how he would improve anything. In the tests, Alice wants to sell something at auction, so she creates the auction instance and (keeping the auction reference to herself at first) puts the item up for auction. Then, only when she is assured that the item is up at auction does she spread the auction instance around to potential bidders. 

There's still the question of how, without invites, the potential bidders find the `zoeInstance` credible. That needs more work.

core/zoe/docs/SecondPriceAuction.md
```diff
@@ -0,0 +1,81 @@
+# Second-price auction
+
+In a second-price auction, the winner is the participant with the
+highest bid, but the winner only pays the price corresponding to the
+second highest bid. Second-price auctions must be sealed to have the
```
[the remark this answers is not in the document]
katelynsills: Hmm, I thought this was clear, but I will add clarification.

core/zoe/zoe.chainmail
katelynsills: Good idea. I've added an insist that will ensure it is undefined

core/zoe/docs/RightsInZoe.md
```diff
@@ -0,0 +1,38 @@
...
+
+Various governing contracts may also want to create further rights,
+such as the right to enter a particular contract instance (an invite).
+
+An invite quantity has the following format:
+
+```js
+{
+  id: {},
+  offerToBeMade: [rule1, rule2, rule3, ...],
```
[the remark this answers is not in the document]
katelynsills: I think it depends on whether the governing contract chooses to enforce it or not. For instance, you could imagine an invite to a swap that does have a specific counter-offer that it will accept and nothing else. Or, it could be an auction that will accept any sort of bid over a minimum. 

Since nothing in this PR or autoswap uses invites, I think I should take this file out for this PR. It will probably change with a PR that does include invites.

core/zoe/zoe.chainmail
katelynsills: Yup, this fix was already made in zoe.js, but I missed it in the chainmail. Let me fix it.

core/zoe/zoe.chainmail
katelynsills: Good point

core/zoe/zoe/zoe.js
katelynsills: What kind of pros and cons are you thinking of compared to how it is now?

core/zoe/zoe.chainmail
katelynsills: Ah, the names here aren't quite correct. I just pushed up a fix. The ERTP payment is called `claimPayoff` and the method that is used is called `getPayoff`. The chainmail struct is correct that it is a property. We should probably work on the names

core/zoe/docs/SWAPS.md
```diff
@@ -0,0 +1,78 @@
...
+And then makes an offer using the escrowReceipt and tries to collect her winnings:
+
+```js
+const aliceOfferResultP = simpleSwap.makeOffer(aliceEscrowReceipt);
+const aliceSeat = await aliceClaimPayoff.unwrap();
+const aliceWinningsP = aliceSeat.getPayoff();
+
+```
+
+She then spreads the zoe instance widely, and Bob decides to be the
```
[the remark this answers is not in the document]
katelynsills: Yeah. I think if I move to Dean's idea of not having to pass in issuers at the start, I can get rid of the entire `zoeInstance` concept which should solve this problem

katelynsills:
I went through the comments and I think a couple of things were coming up consistently:

1)  How does Alice create a new auction/swap with Zoe? Is this easy and intuitive?
2) How does Alice credibly invite Bob to participate in the auction/swap? Can this be made easier with particular solutions to step 1?
3) How does Zoe work in a SwingSet environment where the issuers of underlying rights as well as Alice and Bob may all be in different vats? A SwingSet environment will enforce specific ways of reusing code because functions can't be passed. 

I think there may be some low-hanging fruit in addressing these questions, so I think I will try to do it in this PR rather than after it. One thing is Dean's suggestion that issuers could be stored in a map and added at will rather than at the beginning. I think this may get rid of the `zoeInstance` and leave only the long-lived `zoe` and the governing contract.

core/zoe/docs/SWAPS.md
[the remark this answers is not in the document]
erights: > Yeah. I think if I move to Dean's idea of not having to pass in issuers at the start, I can get rid of the entire zoeInstance concept which should solve this problem

That seems unlikely to me. How?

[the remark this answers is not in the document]
katelynsills: I'm working on it right now and it passes the automaticRefund tests. I can push it up tomorrow but here's what it looks like from the user's perspective:

```js
    // 1: Alice creates an automatic refund instance
    const { instance: aliceAutomaticRefund, instanceId } = zoe.makeInstance(
      'automaticRefund',
      issuers,
    );
```

Alice can pass the instanceId to Bob he can make sure this is the contract that he wants to participate in:

```js
    const { instance: bobAutomaticRefund, libraryName } = zoe.getInstance(
      instanceId,
    );
    t.equals(libraryName, 'automaticRefund');
```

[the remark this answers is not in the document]
katelynsills: Oh, and Bob and Alice escrow with the "actual" Zoe. No such thing as a "zoeInstance" exists:

```js
    const {
      escrowReceipt: allegedBobEscrowReceipt,
      claimPayoff: allegedBobClaimPayoff,
    } = await zoe.escrow(bobOfferDesc, bobPayments);
```

katelynsills:
I'm closing this PR in favor of a new set of Zoe PRs in which #123, #125, and #136 are squashed and then pulled apart again.

PR #135, opened 2019-09-24
add getTotalSupply() method

katelynsills:
Our autoswap implementation benefits from being able to find out the current supply of the liquidity token. It uses the current supply to calculate how many liquidity tokens should be minted when liquidity is added, and how much underlying liquidity should be given back when liquidity is removed and the liquidity tokens are burned. 

This PR adds a custom configuration that can be passed to `makeMint`. The custom configuration, if used when making a mint, adds a method to the `mint` object called `getTotalSupply`, which returns an amount representing the total amount in purses and payments. 

The total supply is calculated in a custom `mintKeeper` called the `totalSupplyMintKeeper`. Whenever the amount of a purse or payment is updated, it alters the `totalSupply` variable within the `mintKeeper`. 

I think we could easily extend this in the future to create a mint that is credibly capped to a certain supply.

### DIFF
--- a/core/issuers.js
+++ b/core/issuers.js
@@ -219,13 +219,20 @@ Description must be truthy: ${description}`;
     },
 
     burnExactly(amount, srcPaymentP) {
-      const sinkPurse = coreIssuer.makeEmptyPurse('sink purse');
-      return sinkPurse.depositExactly(amount, srcPaymentP);
+      return Promise.resolve(srcPaymentP).then(srcPayment => {
+        insistAmountEqualsPaymentBalance(amount, srcPayment);
+        const paymentAmount = paymentKeeper.getAmount(srcPayment);
+        paymentKeeper.remove(srcPayment);
+        return paymentAmount;
+      });
     },
 
     burnAll(srcPaymentP) {
-      const sinkPurse = coreIssuer.makeEmptyPurse('sink purse');
-      return sinkPurse.depositAll(srcPaymentP);
+      return Promise.resolve(srcPaymentP).then(srcPayment => {
+        const amount = paymentKeeper.getAmount(srcPayment);
+        paymentKeeper.remove(srcPayment);
+        return amount;
+      });
     },
   });
 
--- a/test/unitTests/core/test-totalSupplyMintKeeper.js
+++ b/test/unitTests/core/test-totalSupplyMintKeeper.js
@@ -1,0 +1,29 @@
+import { test } from 'tape-promise/tape';
+import { makeMint } from '../../../core/issuers';
+import { makeTotalSupplyConfig } from '../../../core/config/totalSupplyConfig';
+
+test('totalSupplyConfig', async t => {
+  try {
+    const mint = makeMint('liquidityTokens', makeTotalSupplyConfig);
+    const issuer = mint.getIssuer();
+    const purse1 = await mint.mint(1000);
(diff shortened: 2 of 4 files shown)


core/issuers.js
```diff
@@ -207,14 +208,29 @@ Description must be truthy: ${description}`;
       });
     },
 
-    burnExactly(amount, srcPaymentP) {
+    burnExactly: (amount, srcPaymentP) => {
```
Chris-Hibbert: Why convert these two functions to this syntax?

core/issuers.js
```diff
@@ -217,14 +218,29 @@ Description must be truthy: ${description}`;
...
       const sinkPurse = coreIssuer.makeEmptyPurse('sink purse');
-      return sinkPurse.depositExactly(amount, srcPaymentP);
+      return sinkPurse
+        .depositExactly(amount, srcPaymentP)
+        .then(amountBurned => {
```
Chris-Hibbert: This adds an extra turn to the response. What would you think of 

```
         if (!hasTotalSupply()) {
            return sinkPurse.depositExactly(amount, srcPaymentP);
          } else {
            sinkPurse
              .depositExactly(amount, srcPaymentP)
              .then(amountBurned => {
                 totalSupply = assay.without(totalSupply, amountBurned);
              }
             return amountBurned;
          }
```

katelynsills:
> mintKeepers use a weakMap explicitly so that when someone drops a purse/payment, the keeper won't hold onto it. So the totalSupply is only an upper bound. Does this approach to tracking totalSupply actually solve autoSwap's problem?

Ahhh. Good point. What would an alternative look like, since we can't iterate over the weakMap?

katelynsills:
Here's another idea: a custom mintKeeper can keep track of the total supply. It's already part of the configuration, so we wouldn't need to add 'hasTotalSupply'. Hmm. I will try this.

Chris-Hibbert:
I like that idea.

katelynsills:
Ok, I think this version is much cleaner. thanks for the very useful feedback! I am editing the description to better relate what it is now

test/unitTests/core/test-totalSupplyMintKeeper.js
```diff
@@ -0,0 +1,29 @@
+import { test } from 'tape-promise/tape';
+import { makeMint } from '../../../core/issuers';
+import { makeTotalSupplyConfig } from '../../../core/config/totalSupplyConfig';
+
+test('totalSupplyConfig', async t => {
```
Chris-Hibbert: We could simulate dropping a payment by calling split in a separate function, or at least a sub-scope, and letting part of it go out of scope. 

I realize that it won't change totalSupply, but it's still good to have that case explicitly represented, and show what we expect.

katelynsills: Ok, I can simulate dropping the payment, and can show that the totalSupply doesn't change.  Since we can't know when garbage collection will happen, we can't include that part in the test, but it sounds like that's ok.

katelynsills:
Closed in favor of a future new PR for branch [get-total-supply-2](https://github.com/Agoric/ERTP/tree/get-total-supply-2) which is based on master and is not part of the Zoe stack of PRs. It does currently require the generator function version of the configs to work.

PR #138, opened 2019-09-25
WIP: Renaming Parts of ERTP

katelynsills:
THIS PR IS FOR COMMENTING ON THE NAMING. PLEASE COMMENT HERE WHERE APPROPRIATE.

Please see Issue #137 for the source of truth for the renaming.

This is a work in progress so that we can see it before we get too far. Only issuers.js and assay.js have been changed. It's possible that a mistake was made because no tests have been changed yet. 

* This should not be merged in its current state *


katelynsills:
Great! I'll submit a real PR for master and then change the zoe-merged branch

PR #139, opened 2019-09-26
Split Payments Bug Fix and Test

katelynsills:
I talked to Mark about the bug that we were experiencing, in which a `vouch(amount)` failed, specifically for `split`. 

It turns out this is entirely expected behavior, since amounts are pass-by-copy. When the amount gets transmitted across a vat boundary, it gets a new object identity every time. So, if Vat A sends back the amount that it got from Vat B, Vat B will not recognize it and a `vouch` without a `coerce` will fail. 

After talking to Mark further, we realized that `vouch` actually isn't needed at all, and we can use `coerce`.

This PR removes `vouch`, adds `coerce` to `split` to match the pattern of other mint/issuer methods, and adds a test for the particular bug that we were experiencing in `split`.


katelynsills:
Mark verbally approved this. Merging

PR #145, opened 2019-10-04
WIP: Use Import Manager for Strategies

katelynsills:
CURRENTLY INSECURE. DO NOT MERGE.

This is a PR that I'd like some help with and that probably needs discussion. @Chris-Hibbert @erights  

We had previously discussed using the import manager for strategies. That approach works great for strategies that do not take in parameters. For instance, the Nat strategy is an object that doesn't allow for configuration of the strategy, and the Nat strategy works great with this approach. 

However, other strategies have relied on being able to pass in custom functionality in the form of passing in parameters. It's not clear how this should be handled. Here are two options:

1) No custom functionality is allowed in the strategies that are put in the import manager. If custom functionality is desired, different strategies for each possibility must be made and added to the import manager library. For instance, the `pixelStrategy` takes in a parameter `canvasSize` that represents the number of pixels across and down the square canvas. We can't pass in `canvasSize`, but we can hard-code desirable options like so: `pixelStrategy10: makePixelStrategy(10)`

2) We change `assay.js` to pass in parameters to the strategy. So in line 27 of `assay.js`, when we retrieve the strategy from the `strategyLib`, we would instead get back a `makeStrategy` method that we would call with parameters that are passed in from `issuers.js`.

Option 2 seems much more plausible, but it also seems like some work and revision of things that we thought were settled in ERTP. 

For right now, this PR tries to make due with Option 1, but there are consequences. In addition to the hard-coding of things like `pixelStrategy10`, we can't use the `descriptionCoercer` in `contractHost.js` that looks up descriptions in the `seatDescriptions` weakMap. So that part is excised, but we are losing an additional check by taking it out.


core/config/strategies/uniStrategy.js
```diff
@@ -7,64 +7,58 @@ import {
...
-      insist(
-        !!optDescription,
-      )`Uni optDescription must be either null or truthy: ${optDescription}`;
-      mustBeComparable(optDescription);
-      const description = descriptionCoercer(optDescription);
-      insist(!!description)`Uni description must be truthy ${description}`;
-      mustBeComparable(description);
-      return description;
+      mustBeComparable(quantity);
+      return customInsistKind(quantity);
```
erights: The name `customInitKind` suggests that it just approves or rejects. However, here you're returning what it returns, giving it the power to coerce. Which should it be?

core/config/strategies/uniStrategy.js
```diff
@@ -7,64 +7,58 @@ import {
...
+      // left and right can only be added together if one of them is null
+      if (uniStrategy.isEmpty(left)) {
         return right;
       }
-      if (right === null) {
+      if (uniStrategy.isEmpty(right)) {
         return left;
       }
-      if (sameStructure(left, right)) {
-        // The "throw" is useless since insist(false) will unconditionally
```
erights: Please preserve the explanatory comment about the throw in `throw insist(false)...;`

core/config/strategies/uniStrategy.js
```diff
@@ -7,64 +7,58 @@ import {
...
-        throw insist(
-          false,
-        )`Cannot combine different uni descriptions ${left} vs ${right}`;
-      }
+      throw insist(false)`Cannot combine uni quantities ${left} and ${right}`;
     },
     without: (whole, part) => {
-      if (part === null) {
+      // we can only subtract the part from the whole if either part
+      // is null or the part equals the whole
```
erights: "or the part equals the whole" what? And this doesn't correspond to the code.

core/config/strategies/uniStrategy.js
```diff
@@ -7,64 +7,58 @@ import {
...
+      if (uniStrategy.isEmpty(part)) {
         return whole;
       }
-      insist(whole !== null)`Empty left does not include ${part}`;
```
erights: Why did this error check go away? Why would `without` be correct without it?

core/config/strategyLib.js
```diff
@@ -0,0 +1,31 @@
...
+import { makeUniStrategy } from './strategies/uniStrategy';
+import { makePixelStrategy } from '../../more/pixels/pixelStrategy';
+
+import { insist } from '../../util/insist';
+import { mustBeComparable } from '../../util/sameStructure';
+
+const makeCustomInsistKind = (
+  descriptionCoercer = () => true,
+) => optDescription => {
+  insist(
```
erights: The purpose of customInsistKind seems to be to allow these checks around `descriptionCoercer` to be configurable. I doubt this is a good idea. I think these extra checks should be universal and only the remaining `descriptionCoercer` behavior should be configurable.

core/config/strategyLib.js
```diff
@@ -0,0 +1,31 @@
...
+    !!optDescription,
+  )`Uni optDescription must be either null or truthy: ${optDescription}`;
+  const description = descriptionCoercer(optDescription);
+  insist(!!description)`Uni description must be truthy ${description}`;
+  mustBeComparable(description);
+  return description;
+};
+
+const customInsistKind = makeCustomInsistKind();
```
erights: This is where we lose the description checking mentioned in the PR comment? The PR comment suggests losing this is only losing a belt-and-suspenders double check. Actually, I'm not sure whether we're even still secure without the check. TODO Need to examine.

core/config/uniConfig.js
```diff
@@ -0,0 +1,14 @@
+import harden from '@agoric/harden';
+
+import { noCustomization } from './noCustomization';
+import { makeCoreMintKeeper } from './coreMintKeeper';
+
+function makeUniConfig() {
+  return harden({
+    ...noCustomization,
+    makeMintKeeper: makeCoreMintKeeper,
+    strategyName: 'uniStrategy',
```
erights: I see the problem here, right?

core/contractHost.js
```diff
@@ -23,15 +23,7 @@ function makeContractHost(E, evaluate) {
   const seatDescriptions = makePrivateName();
   // from installation to source code string
   const installationSources = makePrivateName();
-
-  function descriptionCoercer(allegedDescription) {
-    const seatDesc = seatDescriptions.get(allegedDescription.seatIdentity);
-    mustBeSameStructure(seatDesc, allegedDescription);
```
erights: Yeah, I suspect we can't simply lose this check. But we might be able to refactor it elsewhere.

Does Zoe still use the `contractHost`? If not, where does this issue reappear in Zoe's seat logic?

more/pixels/pixelConfig.js
```diff
@@ -152,7 +151,7 @@ function makePixelConfigMaker(
         });
       },
       makeMintKeeper: makePixelMintKeeper,
-      strategy: makePixelStrategy(canvasSize),
+      strategyName: 'pixelStrategy10',
```
erights: I understand that hardcoding a 10 here is an awful kludge that we want to get away from as soon as possible. But as long as this PR does this hardcoding, it should also do an `insist` that the actual `canvasSize` is indeed 10

Chris-Hibbert:
I think being able to pass in parameters is pretty important. I'd like to explore ways of making that possible.

erights:
> Option 2 seems much more plausible, but it also seems like some work and revision of things that we thought were settled in ERTP.

Suggestion:

Let's divide this into a medium term solution and a short term step towards this medium term solution.

For the medium term solution, I think something like the old Data-E http://wiki.erights.org/wiki/Safe_Serialization_Under_Mutual_Suspicion pass-by-copy serialization is a good model. We would extend our marshalling logic with something like the Data-E `optUncall` method, and make strategy instances as those kinds of pass-by-copy objects. We would stick with the current importManager approach, but eventually with guaranteed integrity of transitive imports. As Kate's (2) choice suggests, the registered names would be strategy makers. The arguments would themselves be pure pass-by-copy arguments, with this extended notion of pass-by-copy. Thus, we'd still avoid the possible long term of general mobile code, and stick with pre-arranged names for agreed mutually vetted code. Applying the same pass-by-copy extension to the arguments should enable us to deal with the behavioral `descriptionCoercer` issue as well.

For the short term step towards this medium term solution, we could invent a notion of a `strategyDesc`, to serve in the role of this PR's change to `strategyName`. A `strategyDesc` would consist of `{ strategyMakerName, strategyMakerArgs }`. The strategyArgs would be simply pass-by-copy by our current pass-by-copy rules.

This would be adequate for our pixel strategy, as `[10]` is simply pass-by-copy. It would not be adequate for the `descriptionCoercer`. Rather, we should do for the description coercer what this PR does for the pixel strategy: we should make and name a specialized seat strategy maker, with the seat description coercer code, parameterized only by the residual data.

katelynsills:
Closing because we decided to squash the Zoe PRs

PR #146, opened 2019-10-04
Offer Safety Helpers Round 2

katelynsills:
This PR is part of the Zoe implementation. It adds helper functions for determining whether a reallocation of quantities conserves rights (has the same total quantity per issuer) and whether a reallocation is "offer-safe" for each user.

Note that this also has the `get-strategy-name` branch changes in it. I couldn't branch off of both `get-strategy-name` and `seat-mint` and I wanted to keep both of those PRs based on  master.


katelynsills:
Closed because we decided to squash the Zoe PRs

PR #147, opened 2019-10-04
Zoe Basic Contracts Round 2

katelynsills:
This PR adds Zoe plus three very basic contracts:

1. Automatic Refund: a trivial contract that gives the user back what they put in.
2. Simple Offer Swap: a swap contract in which there is only one method in the governing contract API: `makeOffer`. The instance id must be passed around to the various parties for them to make their offer. There are no invites to make an offer.
3. Simple Offer Second-price Auction: a second-price auction contract that uses the same "middle layer" as Simple Offer Swap, but makes changes to the logic of when an offer is valid and how the quantities should be reallocated such that it's a second price auction rather than a swap.

AutomaticRefund can be used to explain the basics of Zoe and for testing, but isn't actually useful. 

The Simple Offer Swap and the Simple Offer Second-price Auction are useful and show that we can build reusable code on top of Zoe such that only the "business logic" of the particular institution changes. 

There's [documentation on the swap and second-price auction](https://github.com/Agoric/ERTP/tree/zoe-basic-contracts/core/zoe/docs), but more documentation is needed.

Mark noted that the second price auction in particular depends on the person selling an asset at auction being the first to use it. That part still remains the same since it seems like it will be fixable after we figure out more pressing details of Zoe. So, `simpleOffer` should be viewed as a proof of concept of a middle layer rather then the final version middle layer that we end up with.


katelynsills:
Closed because we decided to squash all of the Zoe PRs together

PR #151, opened 2019-10-05
WIP: Zoe

katelynsills:


core/zoe/contracts/coveredCall.js
```diff
@@ -0,0 +1,131 @@
...
+      const wantedOffers = srcs.makeWantedOfferDescs(offerMadeDesc);
+
+      const invites = wantedOffers.map(async offer =>
+        zoe.makeInvite(offer, harden({ makeOffer: makeOfferMaker(offer) })),
+      );
+
+      const inviteP = invites[0];
+
+      outcomeP.res('offer successfully made');
+      const [outcome, invite] = await Promise.all([outcomeP.p, inviteP]);
```
erights: Equivalent:

```js
const outcome = 'offer successfully made';
const invite = await inviteP;
```

or better, see below

core/zoe/contracts/coveredCall.js
```diff
@@ -0,0 +1,131 @@
...
+       * Seat: the seat for the initial player
+       * Invites: invitations for all of the other seats that can
+       * be sent to other players.
+       * Both seat and invites are ERTP payments that can be
+       * `unwrap`ed to get a use object.
+       */
+      return harden({
+        outcome,
+        invite,
+      });
```
erights: Equivalent:

```js
return harden({
  outcome: 'offer successfully made',
  invite: await inviteP,
});
```

core/zoe/contracts/coveredCallSrcs.js
```diff
@@ -0,0 +1,71 @@
...
+      {
+        rule: firstOffer[0].rule,
+        assetDesc: firstOffer[1].assetDesc,
+      },
+    ]);
+  return harden([makeSecondOffer(firstOfferDesc)]);
+};
+
+const isValidOfferDesc = (extentOps, offerDescToBeMade, offerDescMade) => {
+  const ruleEqual = (leftRule, rightRule) => leftRule.rule === rightRule.rule;
```
erights: I keep getting tripped up by the part-vs-whole "rule" terminology. Is "rule" one of the strings `offerExactly`, etc...

Or is "rule" a record containing one of those strings and an assetDesc? 

I think you mostly mean the record, in which case the string might be "disposition"?

core/zoe/contracts/coveredCallSrcs.js
```diff
@@ -0,0 +1,71 @@
...
+    extentOp.equals(leftRule.assetDesc.extent, rightRule.assetDesc.extent);
+
+  const assayEqual = (leftRule, rightRule) =>
+    leftRule.assetDesc.label.assay === rightRule.assetDesc.label.assay;
+
+  // Check that two offers are equal in both their rules and their assetDescs
+  const offerEqual = (leftOffer, rightOffer) => {
+    const isLengthEqual = leftOffer.length === rightOffer.length;
+    if (!isLengthEqual) {
+      return false;
```
erights: With this line, I just noticed that the srcs APIs indicate failure with a `false` rather than something with diagnostic information. This can become frustrating when something that should have succeeded doesn't.

That's why the rejected promise carries a "reason". For synchronous cases where failure likely indicates that someone has a bug, I've generally used normal termination vs throwing an exception, such as with `insist`.

core/zoe/zoe.chainmail
```diff
@@ -0,0 +1,150 @@
...
+interface ZoeGoverningContractFacet () {
+
+  /** 
+   * Instruct Zoe to try reallocating for the given offerIds.
+   * Reallocation is a matrix (array of arrays) where the rows are the
+   * extents to be paid to the player who made the offer at the same
+   * index in the offerIds array. The reallocation will only happen if
+   * 'offer safety' and conservation of rights are true, as enforced by Zoe.
+   */
+  reallocate (instanceId :Object, offerIds :List(Object), reallocation :List(List(Extent)));
```
erights: I thought this was already fixed --- that the facet is specific to an instance, so the methods don't need to carry the instanceId.

core/zoe/zoe/zoe.js
```diff
@@ -0,0 +1,331 @@
...
+        const inviteExtent = harden({
+          id: harden({}),
+          instanceId,
+          offerToBeMade,
+        });
+        const invitePurseP = inviteMint.mint(inviteExtent);
+        inviteAddUseObj(inviteExtent.id, useObj);
+        const invitePaymentP = invitePurseP.withdrawAll();
+        return invitePaymentP;
+      },
```
erights: This is the other side of the higher-order assayability issue (aka the Fred problem). The `inviteExtent` must contain all the information that Fred provides to the generic swapper (aka escrow exchange agent), to ensure that Alice cannot cheat him --- cannot sell him something other than the coveredCall he wants.

For the coveredCall, where would the (timing authority, deadline) go? Since the instanceId is specific to an instance, and Fred can credibly obtain instance info, perhaps we extend instances with terms?

Or perhaps the `offerToBeMade` parameter here is misnamed, and it is more like the contractHost's seatDesc?

katelynsills:
Merging into master for further refinement

PR #153, opened 2019-10-12
feat(contractHost): support new moduleFormat bundling

michaelfig:
Please see this change to make the contractHost compatible with the new `ag-solo upload-contract` bundling mechanism.

### DIFF
--- a/core/contractHost.js
+++ b/core/contractHost.js
@@ -93,9 +93,27 @@ No invites left`;
     // number of functions with names beginning 'check', each of which can be
     // used by clients to help validate that they have terms that match the
     // contract.
-    install(contractSrcs) {
-      const installation = extractCheckFunctions(contractSrcs);
-      const startFn = evaluateStringToFn(contractSrcs.start);
+    install(contractSrcs, moduleFormat = 'object') {
+      let installation;
+      let startFn;
+      if (moduleFormat === 'object') {
+        installation = extractCheckFunctions(contractSrcs);
+        startFn = evaluateStringToFn(contractSrcs.start);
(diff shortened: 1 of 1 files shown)


katelynsills:
Hi Michael, I have some questions. When would we use the two `moduleFormat`s? What is each intended for?  

It looks like some of the code in the `moduleFormat === 'getExport'` case is similar to the code in `extractCheckFunction`. Is that on purpose?

michaelfig:
> Hi Michael, I have some questions. When would we use the two `moduleFormat`s? What is each intended for?

The default `moduleFormat`, `"object"` is to maintain compatibility with current uses (if there are any).
 
> It looks like some of the code in the `moduleFormat === 'getExport'` case is similar to the code in `extractCheckFunction`. Is that on purpose?

Yes, the check functions are extracted from the exported namespace, but they are already functions, not source strings.

erights:
Since the new moduleFormat is specific to rollup, should we give it a less generic name than `getExport`?

erights:
Does something test this new case?

erights:
Nevermind. I see https://github.com/Agoric/cosmic-swingset/pull/91

michaelfig:
> Since the new moduleFormat is specific to rollup, should we give it a less generic name than `getExport`?

Actually, the `getExport()` wrapping of the rollup code is something that originated with SwingSet, and theoretically could apply also to non-rollup bundlers, too.  It's for any code for which:

```js
const moduleNamespace = getExport();
```

So I'm inclined to leave it as-is.

Thanks!

PR #154, opened 2019-10-13
Registrar service

dtribble:
~~PR FOR DISCUSSION ONLY~~

This illustrates the registrar naming service. You register an object with a key prefix, and it returns the key, of the form `<prefix>_<digits>`. The `digits` are from an integer in a sparse distribution. the distribution is deterministic, but this helps protect against typos resulting in generating another valid key.

```
command[0] home.registrar~.register('handoff', home.handoffService)
history[0] "handoff_4409"

command[1] home.registrar~.get('handoff')
history[1] Promise.reject("Error: Key must end with _<digits> (a string)\nSee console for error data.")

command[2] home.registrar~.get('handoff_4409')
history[2] [Presence o-51]

command[3] history[2] === home.handoffService
history[3] true
```

Outstanding issues:
- ~~no tests~~
- no iteration (may not be required)
- it uses `get` rather than simple property lookup (which requires a proxy)

### DIFF
--- a/more/registrar/registrar.js
+++ b/more/registrar/registrar.js
@@ -1,15 +1,17 @@
 // Copyright (C) 2019 Agoric, under Apache License 2.0
 
-import harden from "@agoric/harden";
-import { insist } from "../../util/insist";
-import { generateSparseInts } from "../../util/sparseInts";
+import harden from '@agoric/harden';
+import { insist } from '../../util/insist';
+import { generateSparseInts } from '../../util/sparseInts';
 
 // Example REPL imnvocations
 // home.registrar~.register('handoff', home.handoffService)
 // for (let i = 0; i < 3000; i++) { home.registrar~.register('handoff', home.handoffService); }
 
+const minimumDigits = 4;
+
 // Generated keys must end with _ and digits.
-const keyFormat = new RegExp('.*_[0-9]+');
+const keyFormat = new RegExp(`.*_\\d{${minimumDigits},}$`);
 
 function makeRegistrar(systemVersion, seed = 0) {
   // TODO make a better token algorithm.
@@ -23,7 +25,10 @@ function makeRegistrar(systemVersion, seed = 0) {
       const realName = name.toLowerCase();
       const useCount = (useCounts.get(realName) || 0) + 1;
       useCounts.set(realName, useCount);
-      const depth = Math.max(4, Math.floor(Math.log10(useCount) + 1.6));
+      const depth = Math.max(
+        minimumDigits,
+        Math.floor(Math.log10(useCount) + 1.6),
+      );
 
       // Retry until we have a unique key.
       let key;
@@ -34,14 +39,14 @@ function makeRegistrar(systemVersion, seed = 0) {
         key = `${realName}_${keyString}`;
       } while (contents.has(key));
 
-      contents.set(key, value);
+      contents.set(key, harden(value));
       return key;
     },
     get(key, version = null) {
       insist(typeof key === 'string')`\
 Key must be string ${key}`;
       insist(keyFormat.test(key))`\
-Key must end with _<digits> ${key}`
+Key must end with _<digits> ${key}`;
       if (version) {
         insist(version === systemVersion)`\
 Key is from incompatible version: ${version} should be ${systemVersion}`;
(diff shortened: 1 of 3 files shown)


more/registrar/registrar.js
```diff
@@ -0,0 +1,55 @@
...
+      const realName = name.toLowerCase();
+      const useCount = (useCounts.get(realName) || 0) + 1;
+      useCounts.set(realName, useCount);
+      const depth = Math.max(4, Math.floor(Math.log10(useCount) + 1.6));
+      const uniqueString = sparseInts.next().value.toString();
+      const keyString = uniqueString.slice(-depth).padStart(depth, '0');
+      // console.log(`RAND ${useCount} ${uniqueString} ${keyString}`);
+      const key = `${realName}_${keyString}`;
+      // if it was a random keyString, then we need to detect collision
+      insist(!contents.has(key), 'Generated name must not collide');
```
erights: Wrong insist. Should be

```js
        insist(!contents.has(key))`\
Generated name must not collide`;
```

except collision shouldn't be an error because you're only using 4 digits. Collisions will happen. Should search for a non-collision instead.

more/registrar/registrar.js
```diff
@@ -0,0 +1,55 @@
...
+      return key;
+    },
+    get(key, version = null) {
+      insist(typeof key === 'string')`\
+Key must be string ${key}`;
+      insist(keyFormat.test(key))`\
+Key must end with _<digits> ${key}`
+      if (version) {
+        insist(version === systemVersion)`\
+Key is from incompatible version: ${version} should be ${systemVersion}`;
```
erights: This is an interesting safety check. How do you decide which cases warrant them? AFAICT, the issue seems pervasive so they'd need to go everywhere or nowhere.

In particular, a key from an earlier version will not be found in the map anyway if it shouldn't be. So this seems one of the lowest priority places to add such a check.

more/registrar/registrar.js
```diff
@@ -0,0 +1,55 @@
...
+      }
+      return contents.get(key);
+    },
+    keys() {
+      return contents.keys();
```
erights: This is where the remote iterator issue that you mentioned in wire comes up. For present SFBW purposes, just flatten it into an array, but TODO or file an issue that scaling will demand something else.

```js
return [...contents.keys()];
```

more/registrar/registrar.js
```diff
@@ -0,0 +1,55 @@
...
+import harden from "@agoric/harden";
+import { insist } from "../../util/insist";
+import { generateSparseInts } from "../../util/sparseInts";
+
+// Example REPL imnvocations
+// home.registrar~.register('handoff', home.handoffService)
+// for (let i = 0; i < 3000; i++) { home.registrar~.register('handoff', home.handoffService); }
+
+// Generated keys must end with _ and digits.
+const keyFormat = new RegExp('.*_[0-9]+');
```
Chris-Hibbert: Do you want to say '.*_[0-9]{4,}' to mean at least 4 digits?

more/registrar/registrar.js
michaelfig: Not sure what to do here.  I'm leaving it as-is for now.

more/registrar/registrar.js
michaelfig: Fixed the insist, and added a loop to find a non-colliding key (but should be rare, as @dtribble explained the number of digits is expanded as a name is used).

michaelfig:
Unit tests have been added and comments addressed (except `systemVersion`).

test/unitTests/more/registrar/test-registrar.js
```diff
@@ -0,0 +1,60 @@
...
+
+    t.equals(registrarService.keys().length, 4, 'number of keys is expected');
+  } catch (e) {
+    t.isNot(e, e, 'unexpected exception');
+  } finally {
+    t.end();
+  }
+});
+
+test.only('Registrar collisions', async t => {
```
katelynsills: should this test.only be taken out?

more/registrar/registrar.js
```diff
@@ -0,0 +1,64 @@
...
+// for (let i = 0; i < 3000; i++) { home.registrar~.register('handoff', home.handoffService); }
+
+const minimumDigits = 4;
+
+// Generated keys must end with _ and digits.
+const keyFormat = new RegExp(`.*_\\d{${minimumDigits},}$`);
+
+function makeRegistrar(systemVersion, seed = 0) {
+  // TODO make a better token algorithm.
+  const useCounts = new Map();
```
katelynsills: What is useCounts used for?

PR #155, opened 2019-10-15
feat(zoe): allow specifying additional endowments for contracts

michaelfig:
This is also added to the contractHost, so that SwingSet users
can endow a `require` function for bundled contract code in the
`getExport` module format.


michaelfig:
> In our full module system, we do want to support resource endowments. But only when we can also hand these out only in a least authority manner. Since we're not doing that yet, we should document that nothing should appear here other than pure (or pretend pure).

Okay, I've added these docs.

PR #161, opened 2019-10-16
Simple Exchange

katelynsills:


core/zoe/contracts/simpleExchange.js
```diff
@@ -0,0 +1,155 @@
...
+    addOrder: async escrowReceipt => {
+      const { id, offerMade: offerMadeDesc } = await zoe.burnEscrowReceipt(
+        escrowReceipt,
+      );
+
+      // TODO: check assays as well
+      const isValidSellOrder = newOfferDesc =>
+        ['offerExactly', 'wantAtLeast'].every(
+          (rule, i) => rule === newOfferDesc[i].rule,
+          true,
```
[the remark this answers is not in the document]
katelynsills: whoops, yes it is

core/zoe/contracts/simpleExchange.js
```diff
@@ -0,0 +1,133 @@
...
+      // Eject if the offer is invalid
+      if (!isValidSellOrder(offerMadeDesc) && !isValidBuyOrder(offerMadeDesc)) {
+        zoe.complete(harden([id]));
+        return Promise.reject(
+          new Error(`The offer was invalid. Please check your refund.`),
+        );
+      }
+
+      // Save the valid offer
+      const isSellOrder = offerMadeDesc[0].rule === 'offerExactly';
```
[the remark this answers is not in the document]
katelynsills: good idea, that is neater. I usually like to follow a pattern of putting the failure case first, but I think your last example is nice and clean

core/zoe/contracts/simpleExchange.js
```diff
@@ -0,0 +1,134 @@
...
+// the second asset received in a sell order may be greater than
+// expected, and the amount of the second asset paid in a buy order
+// may be less than expected. This simple exchange does not support
+// partial fills of orders.
+
+const makeContract = harden(zoe => {
+  let sellOrderOfferIds = [];
+  let buyOrderOfferIds = [];
+
+  const offerIdToOrder = new WeakMap();
```
dtribble: I don't think of IDs as having identity; their data is the identity. Thus, I am surprised to see a WeakMap from Id.

core/zoe/contracts/simpleExchange.js
```diff
@@ -0,0 +1,134 @@
...
+        );
+        return assetEqual && priceOk;
+      };
+
+      const extentOps = zoe.getExtentOps();
+
+      const reallocate = (sellOrderOfferId, buyOrderOfferId) => {
+        const offerIds = harden([sellOrderOfferId, buyOrderOfferId]);
+
+        const [sellOrderExtents, buyOrderExtents] = zoe.getExtentsFor(offerIds);
```
dtribble: why are these called "orders" instead of "offers"?

core/zoe/contracts/simpleExchange.js
```diff
@@ -0,0 +1,134 @@
...
+
+        // If there is a difference in what the seller will accept at
+        // least and what the buyer will pay at most, we will award
+        // the difference to the seller for no reason other than
+        // simplicity. Note that to split the difference requires the
```
dtribble: This works for demo purposes. Is it as simple to favor the offer from the order book?

core/zoe/contracts/simpleExchange.js
```diff
@@ -0,0 +1,134 @@
...
+  let buyOrderOfferIds = [];
+
+  const offerIdToOrder = new WeakMap();
+  const getBuyOrders = () => buyOrderOfferIds.map(id => offerIdToOrder.get(id));
+  const getSellOrders = () =>
+    sellOrderOfferIds.map(id => offerIdToOrder.get(id));
+
+  return harden({
+    addOrder: async escrowReceipt => {
+      const { id, offerMade: offerMadeDesc } = await zoe.burnEscrowReceipt(
```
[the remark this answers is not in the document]
katelynsills: Oh, what you mean by "usually"? Assays have a `burnAll` method, which is what this uses, so I think the name is apt. Also, `burnAll` does return the extent, as this methods does. So I think it usually it works as it does here.

core/zoe/contracts/simpleExchange.js
```diff
@@ -0,0 +1,134 @@
...
+        escrowReceipt,
+      );
+
+      const canMatch = (extentOps, sellOrder, buyOrder) => {
```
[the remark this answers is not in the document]
katelynsills: We could move this out to line 22 if we want to (I like that idea), but we can't move it further out given that we can't import modules yet.

core/zoe/contracts/simpleExchange.js
```diff
@@ -0,0 +1,134 @@
...
+          buyOrder[0].assetDesc.extent,
+        );
+        const priceOk = extentOps[1].includes(
+          buyOrder[1].assetDesc.extent,
+          sellOrder[1].assetDesc.extent,
+        );
+        return assetEqual && priceOk;
+      };
+
+      const extentOps = zoe.getExtentOps();
```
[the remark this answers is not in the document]
katelynsills: Good point, this method should be renamed. It's getting an array of the extentOps for this particular governing contract instance.

core/zoe/contracts/simpleExchange.js
```diff
@@ -0,0 +1,134 @@
...
+        );
+        buyOrderOfferIds = buyOrderOfferIds.filter(
+          offerId => offerId !== buyOrderOfferId,
+        );
+        offerIdToOrder.delete(sellOrderOfferId);
+        offerIdToOrder.delete(buyOrderOfferId);
+      };
+
+      // TODO: check assays as well
+      const isValidSellOrder = newOfferDesc =>
```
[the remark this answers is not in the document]
katelynsills: In a prior code review, I used equivalence checks and it was suggested that I use this style, so it sounds like it might be a matter of preference

core/zoe/contracts/simpleExchange.js
katelynsills: I think there might be a misunderstanding here. The offer`id` is an empty unique object that the governing contract needs to keep around because it's how the governing contract talks to Zoe about offers, so to me it makes perfect sense to store the (`id`, `offerDesc`) in a WeakMap as the (`key`, `value`) respectively. 

We might get multiple offers that have the same `offerDesc`, but we need to be able to distinguish between them, so I'm not sure how we can use the `offerDesc` as the identity or why we would want to.

core/zoe/contracts/simpleExchange.js
katelynsills: They could be called offers. I was calling them orders to show writing from an exchange mindset, but that could be not as helpful as I thought.

core/zoe/contracts/simpleExchange.js
katelynsills: Yup, just as simple.

core/zoe/contracts/simpleExchange.js
katelynsills: Would you prefer I call them offers and leave out the word `order`?

core/zoe/contracts/simpleExchange.js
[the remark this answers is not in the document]
katelynsills: Ok, I'm changing this to `zoe.getExtentOpsArray()`. More verbose but it clarifies that it returns an array rather than a single `extentOps`. `zoe` here is the governing contract facet of Zoe, which knows which instance it is.

core/zoe/contracts/simpleExchange.js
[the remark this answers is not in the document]
katelynsills: These functions have been moved out to line 22

core/zoe/contracts/simpleExchange.js
katelynsills: I left out the word `order`. It's clearer

core/zoe/contracts/simpleExchange.js
dtribble: What I mean is that in most systems and contexts, an "id" is a data value that can e.g., be directly marshaled. If it has actual identity, I think it needs to be called a "handle" or "ref" or some such. I find "id" to be actively misleading.

katelynsills: Ah, gotcha. Thanks for explaining.

PR #176, opened 2019-10-22
ES module Zoe contract with tests

michaelfig:
Hi, here is some code to review, for how I used `@agoric/bundle-source` with the publicSwapESM contract.

I wouldn't expect that this would be merged verbatim, but we can use this PR to discuss the approach.


katelynsills:
I'll play around with this more tomorrow, but just looking at the code, this looks really good to me!

katelynsills:
Copied and used in #180. thanks so much!

PR #184, opened 2019-10-24
Initial write up of Zoe

katelynsills:
Initial write up of Zoe plus some changes to the contracts as a result of making things easier to understand for the writeup.

### DIFF
--- a/core/zoe/docs/zoe.md
+++ b/core/zoe/docs/zoe.md
@@ -1,0 +1,263 @@
+# Zoe: Offer-Safety Enforcement
+
+Note: Zoe is currently at the pre-alpha stage. It has not yet been
+formally tested or hardened.
+
+This guide assumes some knowledge of the [ERTP fundamentals](./README.md).
+
+## What is Zoe? 
+
+__For users__: Zoe guarantees that as a user of a smart contract, you
+will either get what you wanted or get a refund back, even if the
+smart contract is buggy or backdoored. (In fact, the smart contract
+never has access to your digital assets.)
+
+__For developers__: Zoe provides a safety net so you can focus on what
+your smart contract does best, without worrying about your users
+losing their assets due to a bug in the code that you wrote. Writing a
+smart contract on Zoe is easy: all of the Zoe smart contracts are
+written in the familiar language of JavaScript. 
+
+## Sounds like magic. How does it actually work?
+
+To use Zoe, we have to put things in terms of "offers". An offer is a
+statement about what you want and what you're willing to offer. It
+turns out, many smart contracts (apart from gifts and one-way
+payments) involve an exchange of digital assets that can be put in
+terms of offers. 
+
+In this version of Zoe, our offers are simple (see [our roadmap](TODO) for
+more complex offer types and their expected release). We can say
+things like, "I'll give you [three
+wood for two bricks](https://en.wikipedia.org/wiki/Catan)." We can
+also say something like, "I'll give you three wood, and *the most* I'm
+willing to pay is two bricks." Or even: "I can pay you two bricks and
+I expect *at least* three wood back." [Learn more about the particulars
+of structuring an offer here](TODO). 
+
+Offers are a structured way of describing user intent. To a certain
+extent, an offer description is the user's *contractual understanding*
+of the agreement they are entering into. You might have noticed that
+the offer doesn't specify the mechanism by which the exchange happens.
+The offer doesn't say whether the item you want is up for auction, in
+an exchange, or part of a private trade. The offer doesn't mention the
+particular mechanism because an important part of the design of Zoe is
+a __separation of concerns__. Zoe is responsible for enforcing what we
+call "offer-safety", and the smart contract that runs on top of Zoe is
+responsible for figuring out a proposed reallocation of resources. To
+use an auction as an example, the smart contract is responsible for
+figuring out who wins the auction and how much they pay, but Zoe
+handles the escrowing of the bids and the payments. You can think of
+this as similar to e-commerce websites using a separate
+payment-processor so that they don't have to handle the credit cards
+themselves.
+
+### What is "offer-safety"?
+
+Zoe guarantees offer-safety, meaning that when a user makes an offer
+that is escrowed with Zoe, Zoe guarantees that the user will either
+get what they said they wanted, or the user will get what they
+originally offered back. 
+
+When a user escrows with Zoe, they get two things back immediately: an escrow
+receipt, and a JavaScript promise for a future payout. This escrow
+receipt is what the user can send to smart contracts, as proof that they
+have escrowed the underlying digital assets with Zoe, without the
+smart contract ever having access to the underlying digital assets.
+Let's look a particular example to see how this works.
+
+## An example: A swap
+
+I want to trade my three bricks for five wool. You realize you have
+five wool and agree to the deal. Without Zoe, though, you might send
+me the five wool, and I might disappear without ever giving you the
+three bricks in return. With Zoe, we can safely trade with each
+other, even if we don't trust one another. We can also safely rely on
+the swap smart contract. At worst, if the swap contract behaves badly,
+we will both get a refund, and at best, we'll get what we each wanted.
+
+Let's use the basic `publicSwap` contract ([full text of
+the real contract](/core/zoe/contracts/publicSwap.js)). 
+
+Here's a high-level overview of what would happen:
+1. I make an instance of the swap contract.
+2. I escrow my three bricks with Zoe and get an escrow receipt and a
+   promise for a payout in return.
+3. I send my escrow receipt to the swap as the first offer.
+4. I tell you the swap's `instanceId`
+5. Using the `instanceId`, you look up the swap with Zoe.
+6. You verify that it's using the `publicSwap` contract
+   code you expect, and can ask the swap about the offers made so far.
+7. You escrow your offer (offering five wool for three bricks) with
+   Zoe, getting an escrow receipt and a promise for a payout in
+   return.
+8. You send your escrow receipt to the swap as a matching offer.
+9. The offer matches and both of our payout promises resolve to [ERTP
+   payments](TODO), mine to the five wool that I wanted, and yours to
+   the three bricks that you wanted. Success!
+
+
+## How to write smart contracts
+
+Writing smart contracts that run on Zoe is easy, but let's look
+at a simple contract. This contract only does one thing, and
+it's pretty useless - it gives you back what you put in. Let's call it
+`automaticRefund`. Let's say the code of `automaticRefund` looks like this (see
+the [real contract code here](/core/zoe/contracts/automaticRefund.js)):
+
+```js
+export const makeContract = (zoe, terms) => {
+  return {
+    instance: {
+      makeOffer: async escrowReceipt => {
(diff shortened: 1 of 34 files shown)


core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+
+## What is Zoe? 
+
+__For users__: Zoe guarantees that as a user of a smart contract, you
+will either get what you wanted or get a refund back, even if the
+smart contract is buggy or backdoored. (In fact, the smart contract
+never has access to your digital assets.)
+
+__For developers__: Zoe provides a safety net so you can focus on what
+your smart contract does best, without worrying about your users
```
Chris-Hibbert: I would say "you can focus on what is central to your contract". "Focus on what you do best" makes sense when we know that the listener has an established position that they want to leverage.

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+
+Let's use the basic `publicSwap` contract ([full text of
+the real contract](/core/zoe/contracts/publicSwap.js)). 
+
+Here's a high-level overview of what would happen:
+1. I make an instance of the swap contract.
+2. I escrow my three bricks with Zoe and get an escrow receipt and a
+   promise for a payout in return.
+3. I send my escrow receipt to the swap as the first offer.
+4. I tell you the swap's `instanceId`
```
Chris-Hibbert: Can't we share the actual object identity using the handoff service or equivalent? "instanceId" makes it sound guessable and definitely not like a capability.

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+complete the offer and send the user their payout. 
+
+A smart contract on Zoe must export a function `makeContract` that
+takes two parameters: `zoe`, which is the governing contract facet of
+Zoe, and `terms`, which are the contract terms that a contract
+instance is made with. `Terms` must include a property called
+`assays`, which is an array of assays, the public faces of mints. For
+instance, in our bricks-for-wool example above, the contract terms
+would include the brick assay and the wool assay. `Terms` can also
+include any other contract-specific parameters that a user wants to
```
Chris-Hibbert: "`Terms` would also include any other contract-specific parameters that the author specified"

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+pass in. 
+
+The smart contract must return an object with two properties:
+`instance`, which is the user-facing API of the
+contract, and `assays`, which is what the contract has decided is the
+canonical list of assays for the contract. If no change are necessary,
```
Chris-Hibbert: s/are/is/

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+`assays` may just be the assays in the terms. 
+
+To get a better idea of the usual control flow, let's look at a more
```
Chris-Hibbert: A new section here would be helpful.

core/zoe/docs/zoe.md
katelynsills: `InstanceId` is a unique object in the same way as offerId, so we can use whatever name we choose for this pattern here too

core/zoe/docs/zoe.md
katelynsills: hmm, I think you're right that it's presumptive. I think it's snappier than the alternative, though

core/zoe/docs/zoe.md
katelynsills: Good point

core/zoe/docs/zoe.md
katelynsills: Good catch

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+Note: Zoe is currently at the pre-alpha stage. It has not yet been
+formally tested or hardened.
+
+This guide assumes some knowledge of the [ERTP fundamentals](./README.md).
+
+## What is Zoe? 
+
+__For users__: Zoe guarantees that as a user of a smart contract, you
+will either get what you wanted or get a refund back, even if the
+smart contract is buggy or backdoored. (In fact, the smart contract
```
erights: "backdoored" is a strange choice. I normally say "buggy or malicious". By "backdoored" I assume there is an illicit communications channel back to the attacker. Anything backdoored is malicious, but many malicious things are not backdoored. Confinement prevents backdoors but, obviously, not malice.

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+never has access to your digital assets.)
+
+__For developers__: Zoe provides a safety net so you can focus on what
+your smart contract does best, without worrying about your users
+losing their assets due to a bug in the code that you wrote. Writing a
+smart contract on Zoe is easy: all of the Zoe smart contracts are
```
erights: "on" vs "in"?

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+handles the escrowing of the bids and the payments. You can think of
+this as similar to e-commerce websites using a separate
+payment-processor so that they don't have to handle the credit cards
+themselves.
+
+### What is "offer-safety"?
+
+Zoe guarantees offer-safety, meaning that when a user makes an offer
+that is escrowed with Zoe, Zoe guarantees that the user will either
+get what they said they wanted, or the user will get what they
```
erights: "get what the originally offered back" --> "get back what they originally offered".

Reads better. Needs less mental stack to parse.

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+
+__For developers__: Zoe provides a safety net so you can focus on what
+your smart contract does best, without worrying about your users
+losing their assets due to a bug in the code that you wrote. Writing a
+smart contract on Zoe is easy: all of the Zoe smart contracts are
+written in the familiar language of JavaScript. 
+
+## Sounds like magic. How does it actually work?
+
+To use Zoe, we have to put things in terms of "offers". An offer is a
```
erights: "we have to put" -> "we put"

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+statement about what you want and what you're willing to offer. It
+turns out, many smart contracts (apart from gifts and one-way
+payments) involve an exchange of digital assets that can be put in
+terms of offers. 
+
+In this version of Zoe, our offers are simple (see [our roadmap](TODO) for
+more complex offer types and their expected release). We can say
+things like, "I'll give you [three
+wood for two bricks](https://en.wikipedia.org/wiki/Catan)." We can
```
erights: I don't know the game. But outside the game, "wood" is a plural noun. "three wood" sounds as non-sensical as "three water"

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+also say something like, "I'll give you three wood, and *the most* I'm
+willing to pay is two bricks." Or even: "I can pay you two bricks and
```
erights: Check the parity of your examples. Here, your offering both wood and bricks and not asking for anything in exchange.

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+When a user escrows with Zoe, they get two things back immediately: an escrow
+receipt, and a JavaScript promise for a future payout. This escrow
+receipt is what the user can send to smart contracts, as proof that they
+have escrowed the underlying digital assets with Zoe, without the
+smart contract ever having access to the underlying digital assets.
+Let's look a particular example to see how this works.
+
+## An example: A swap
+
+I want to trade my three bricks for five wool. You realize you have
```
erights: "wool" has the same plural noun problem.

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+five wool and agree to the deal. Without Zoe, though, you might send
+me the five wool, and I might disappear without ever giving you the
+three bricks in return. With Zoe, we can safely trade with each
+other, even if we don't trust one another. We can also safely rely on
+the swap smart contract. At worst, if the swap contract behaves badly,
```
erights: "we can also safely rely on the swap smart contract" confuses me. To rely on X is to take a risk on X's possible bad behavior. The point is that you can rely on zoe for these safety properties *without* relying on the swap contract.

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+
+Here's a high-level overview of what would happen:
+1. I make an instance of the swap contract.
+2. I escrow my three bricks with Zoe and get an escrow receipt and a
+   promise for a payout in return.
+3. I send my escrow receipt to the swap as the first offer.
+4. I tell you the swap's `instanceId`
+5. Using the `instanceId`, you look up the swap with Zoe.
+6. You verify that it's using the `publicSwap` contract
+   code you expect, and can ask the swap about the offers made so far.
```
erights: "and can ask the swap about the offers made so far". This is an anti-pattern. As we discussed, we can support it. But we should not encourage it.

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+(In a real contract, whenever we create a new object or array, we recursively
+deep-freeze it with `@agoric/harden`. You can [learn more about `harden` here](https://github.com/Agoric/harden).)
+
+`automaticRefund` has one method exposed to the user: `makeOffer`.
+`makeOffer` takes in a `escrowReceipt`, and after burning the
+`escrowReceipt` (and thus verifying it as well), it tells Zoe to
+complete the offer and send the user their payout. 
+
+A smart contract on Zoe must export a function `makeContract` that
+takes two parameters: `zoe`, which is the governing contract facet of
```
erights: "facet" might need to be explained. Not sure how common it is.

core/zoe/docs/zoe.md
```diff
@@ -0,0 +1,263 @@
...
+Zoe, and `terms`, which are the contract terms that a contract
+instance is made with. `Terms` must include a property called
+`assays`, which is an array of assays, the public faces of mints. For
```
erights: No suggested change here. But noting how weird this is. To me, the mint is the private/admin face[t] of the assay/issuer.

You're using both "facet" and "face". Pick one.

core/zoe/docs/zoe.md
katelynsills: good point

core/zoe/docs/zoe.md
katelynsills: oh, excellent catch, thanks

core/zoe/docs/zoe.md
katelynsills: yeah, good point

core/zoe/docs/zoe.md
katelynsills: Yeah, hmm. I guess it depends on whether we want to keep the Catan example or not. I'll see what I can do

core/zoe/docs/zoe.md
katelynsills: good point, I'll take that out

PR #208, opened 2019-10-31
explain traits and remove useless returns

erights:
Explains why and how we use generators to express trait combination.

### DIFF
--- a/core/config/configs.md
+++ b/core/config/configs.md
@@ -116,7 +116,7 @@ hand, we do know that there is a genuine need to have slightly
 different purposes expressed.
 
 This is why we've chosen to use the [Trait
-pattern](https://en.wikipedia.org/wiki/Trait_(computer_programming),
+pattern](https://en.wikipedia.org/wiki/Trait_%28computer_programming%29),
 in which custom behavior can be defined and recombined easily.
 Furthermore, if we always combine the methods such that the "core"
 methods are last, they cannot be overridden by the custom methods.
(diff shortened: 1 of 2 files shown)


core/config/configs.md
```diff
@@ -116,35 +116,26 @@ hand, we do know that there is a genuine need to have slightly
 different purposes expressed.
 
 This is why we've chosen to use the [Trait
-pattern](https://en.wikipedia.org/wiki/Trait_(computer_programming)),
+pattern](https://en.wikipedia.org/wiki/Trait_(computer_programming),
```
Chris-Hibbert: There's an open paren for the link "[...](...)", and another for the wikipedia name. The original was correct.

core/config/configs.md
```diff
@@ -116,35 +116,26 @@ hand, we do know that there is a genuine need to have slightly
...
+    unwrap: () => makeUseObjForPayment(assay, payment),
+  });
+}
+```
+
+Above, the `payment` object is the actual payment to which this trait contributes the `unwrap` method. We introduce this peculiar generator pattern to solve this self-reference problem. The `payment` variable refers to the object as a whole. However, the object-as-a-whole is not yet assembled, and will be assembled only by a distinct piece of code written in a distinct scope.
+
+In the generator function pattern shown above, the trait methods of a given layer are defined in the scope of the `payment` variableis bound on the left of the yield. The `payment` variable is bound to the value that the yield returns. However, before that happens, the generator yields the argument to yield, the object containing the trait's methods, to be combined to make the object itself. The pattern above containing `makeMintTraitIter` first runs one step of the trait generartor to obtain the trait methods, combining them into the overall object. Only when the object is complete does it go back to the generator, in order to bind that object to that trait's variable for the object itself, such as `payment`.
```
Chris-Hibbert: something is broken in the vicinity of "varialbleis", and I can't tell what it should be.

core/config/configs.md
erights: Interesting. You are correct. However, I was editing via Atom's Markdown editor with Markdown preview, which got confused by the parens within the URL. Changed to uuencode the parens, which seems to work both places.

core/config/configs.md
erights: Done

PR #221, opened 2019-11-14
Fix typos and UnitOps description

katelynsills:
I was going through issues and realized #174 was a quick fix. I was especially motivated because there's an embarrassing typo in which replacing "AssetDesc" with "Unit" created "Unitss".  

This PR fixes the typos and the Unit/Unit Ops Description.

### DIFF
--- a/core/unitOps.chainmail
+++ b/core/unitOps.chainmail
@@ -2,12 +2,12 @@
  * Units are asset descriptions and take the form of labeled extents. 
  * UnitOps execute the logic of how units are changed when digital 
  * assets are merged, separated, or otherwise manipulated. For
- * instance, a deposit of 2 bucks into a purse that already has 3 bucks
+ * example, a deposit of 2 bucks into a purse that already has 3 bucks
  * gives a new balance of 5 bucks. An empty purse has 0 bucks. UnitOps 
  * relies heavily on ExtentOps, which manipulates the non-labeled
  * portion.
  */
-struct Units (Extent, Label) {
+struct Units (Label, Extent) {
   label :Label;
   extent :Extent;
 }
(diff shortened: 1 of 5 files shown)


core/unitOps.chainmail
```diff
@@ -1,7 +1,11 @@
 /**
- * Unitss are wrappers on extents that have been validated by an UnitOps, and
- * can be verified as having been issued by the UnitOps. They contain their
- * extent and a Label. The label identifies a particular assay.
+ * Units are asset descriptions and take the form of labeled extents. 
+ * UnitOps execute the logic of how units are changed when digital 
+ * assets are merged, separated, or otherwise manipulated. For
+ * instance, a deposit of 2 bucks into a purse that already has 3 bucks
```
erights: A pervasive issue is the use of "for instance". A minor issue, Since we are explaining to object programmers, I always use "for example" instead. When I see "instance", it always takes me an extra moment to think "oh, not that meaning of 'instance' ". Might be a rare reaction, but probably not unique to me.

core/unitOps.chainmail
```diff
@@ -1,7 +1,11 @@
...
+ * gives a new balance of 5 bucks. An empty purse has 0 bucks. UnitOps 
+ * relies heavily on ExtentOps, which manipulates the non-labeled
+ * portion.
  */
 struct Units (Extent, Label) {
```
erights: I'd reverse the order of the type parameters: `struct Units (Label, Extent)`

katelynsills: Ok, sounds good, that's in

core/unitOps.chainmail
katelynsills: Sure, we can try to use "for example" more

Issue #225, opened 2019-11-15
[Zoe] Simplify the user flow of making an offer and interacting with a smart contract instance


katelynsills:
We just had a conversation where Mark outlined an idea for simplifying the user flow of making an offer with Zoe and interacting with a smart contract instance. I'll try to outline the current flow compared to this new idea. (I'm going all the way back to creating the installation for clarity's sake, even though that part remains the same. I also might include things outside of a particular user's perspective, if it seems important enough. For example, the smart contract calling Zoe or vice versa.)

Current flow of Alice "inviting" Bob to interact with a contract without invites:
1. Alice packs the contract using `bundleSource` from `@agoric/bundle-source`
2. Alice calls `zoe.install` with the outcome of `bundleSource` and gets an `installationHandle` back
3. Alice calls `zoe.makeInstance(installationHandle, terms)` and gets back a record with properties such as instanceHandle, and instance (the "front-door" of the contract).
4. Alice hands over the `instanceHandle` to Bob.
5. Bob calls `Zoe.getInstance(instanceHandle)` and gets back a record that includes the instance as a property, thereby getting the 'front-door'.  He also gets credible information such as the installationHandle and terms that he can use to decide whether to join this contract. 
6. Bob escrows with Zoe by calling `Zoe.escrow(offerRules, offerPayments)` and receives a payout promise and an escrowReceipt
7. Bob calls a method on the contract instance he had gotten in step 5 and passes the escrowReceipt as an argument.

Current flow with invites:
1. Alice packs the contract using `bundleSource` from `@agoric/bundle-source`
2. Alice calls `zoe.install` with the outcome of `bundleSource` and gets an `installationHandle` back
3. Alice calls `zoe.makeInstance(installationHandle, terms)` and gets back a record with properties such as instanceHandle, and instance (the "front-door" of the contract).
4. Alice calls a method on the instance that gives her back an invite that she can give to Bob.
5. Alice gives the invite to Bob.
6. Bob inspects the invite and calls `inviteAssay.claimAll(invite)` to validate it. 
7. Bob escrows with Zoe by calling `Zoe.escrow(offerRules, offerPayments)` and receives a payout promise and an escrowReceipt
8. Bob unwraps the invite to get an object with methods (this is the private version of the "front door" object). 
9. Bob calls a method on the object he had gotten in step 8 and passes the escrowReceipt as an argument.

Flow with new idea:
1. Alice packs the contract using `bundleSource` from `@agoric/bundle-source`
[...]

Issue #224, opened 2019-11-14
[Zoe] Simplify the payoutRule kinds to 'want' and 'offer'


katelynsills:
I had previously pushed for having

* offerAtMost
* offerExactly
* wantExactly
* wantAtLeast

as the possible kinds of payoutRules. The reasoning was that getting more than what was expected can ruin strict accounting practices.

But, in trying to describe offer safety to other people, it turned out that the code for enforcing the offer safety invariant simplifies quite a bit by only having 'want' and 'offer' as the kinds. And, we don't currently have good examples of cases in which the "strict accounting" practices would balk at getting more than expected.

So in weighing an easier developer experience vs a use case that doesn't seem relevant yet, I think we should simplify it for now and then revisit when the "strict accounting" use case actually arises.

warner:
this was issue 224 in the old ERTP repo, and part of the "Zoe - immediate future" project

PR #230, opened 2019-11-19
Zoe refactor spike

katelynsills:
This spike is a WIP implementation of the following:

#222 Smart contracts interact with Zoe in terms of units, not extents
#223 Refactor state.js to be in table/record format rather than individual WeakMaps
#225 Use invites for offers to simplify user flow of escrowing with Zoe and making an offer
#224 Simplify the offer rules to 'want' and 'offer'

#224 didn't have full support, so it may not be kept. (This spike was started before I realized that #224 might be controversial.)

At this point, I'm looking for feedback on design and architecture. Not all tests pass yet, and things haven't been cleaned up to the point that they would normally be for a PR.


Chris-Hibbert (commented):
This looks like a good direction. The table format seems like an improvement.

core/zoe/state.js
```diff
@@ -0,0 +1,254 @@
...
+
+// Installation Table
+// Columns: installationHandle | installation
+const makeInstallationTable = () => {
+  // The WeakMap that stores the records
+  const handleToRecord = makePrivateName();
+
+  const installationTable = harden({
+    // TODO validate installation record
+    validate: allegedInstallationRecord => allegedInstallationRecord,
```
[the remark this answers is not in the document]
katelynsills: Good point. What do you think of instead, having it be an insist which either throws or returns true?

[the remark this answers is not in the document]
Chris-Hibbert: I like that.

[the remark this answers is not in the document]
katelynsills: Ok, that change is in

core/zoe/contracts/automaticRefund.js
```diff
@@ -11,18 +11,26 @@ import harden from '@agoric/harden';
...
+    harden({
+      makeOffer: () => {
+        offersCount += 1;
+        zoe.complete(harden([inviteHandle]), terms.assays);
+        return `The offer was accepted`;
+      },
+      makeInvite: () => {
+        const newInviteHandle = harden({});
+        const seat = makeSeat(newInviteHandle);
+        return zoe.makeInvite(seat, newInviteHandle);
```
erights: Does the made Invite contain the inviteHandle? If so, `zoe.makeInvite` should create it rather than taking it as a parameter.

I'm confused by the lack of a `seatDesc` or anything for that role, so that the holder of an allegedInvite can (hopefully typically) look at the contract source and know which call to makeInvite made the invite they are allegedly holding.

core/zoe/contracts/automaticRefund.js
```diff
@@ -11,18 +11,26 @@ import harden from '@agoric/harden';
...
+      },
+      getOffersCount: () => offersCount,
+    });
   return harden({
-    instance: automaticRefund,
+    initialSeat: makeSeat(initialInviteHandle),
+    initialInviteHandle,
```
erights: I'm surprised that you don't return the initial invite. AFAICT, there is no initial invite, in which case it is weird that there is an `initialInviteHandle`.

core/zoe/contracts/automaticRefund.js
erights: I see. You need the seat to make the invite (correct) and you need the seat to know the inviteHandle (which I don't understand yet), so you need the inviteHandle to exist before you make the invite.

However, the seat doesn't actually need the handle until one of its methods are invoked.  Depending on  why the seat needs the handle, you might be able to solve this simply by initializing a const variable from what `makeInvite` returns, where that variable is in scope for the seat code.

core/zoe/contracts/coveredCall.js
```diff
@@ -16,7 +16,7 @@ export const makeContract = harden((zoe, terms) => {
       offerHandle,
       offerRules: { payoutRules },
     } = await zoe.burnEscrowReceipt(escrowReceipt);
```
erights: For any file in which I still see mention of a receipt, I assume that this has not been converted to new style and so should not be reviewed?

core/zoe/contracts/publicAuction.js
```diff
@@ -20,75 +20,111 @@ export const makeContract = harden((zoe, terms) => {
...
-      const {
-        offerHandle,
-        offerRules: { payoutRules },
-      } = await zoe.burnEscrowReceipt(escrowReceipt);
+  const makeBidderSeat = inviteHandle =>
+    harden({
+      bid: () => {
+        // Check that the item is still up for auction
+        const { inactive } = zoe.getOfferStatuses(
+          harden([creatorInviteHandle]),
```
erights: Maybe "seller" instead of "creator"

core/zoe/contracts/helpers/userFlow.js
```diff
@@ -2,11 +2,12 @@ import harden from '@agoric/harden';
...
 export const rejectOffer = (
   zoe,
-  offerHandle,
+  assays,
+  inviteHandle,
   message = `The offer was invalid. Please check your refund.`,
 ) => {
-  zoe.complete(harden([offerHandle]));
-  return Promise.reject(new Error(`${message}`));
+  zoe.complete(harden([inviteHandle]), assays);
```
erights: Why does `complete` need an `assays` array? Doesn't it already know this according to the per-contract facet it is?

core/zoe/contracts/publicAuction.js
```diff
@@ -20,75 +20,111 @@ export const makeContract = harden((zoe, terms) => {
...
+        );
+        if (inactive.length > 0) {
+          return rejectOffer(
```
erights: Since `rejectOffer` itself only throws, it is arguably better that you should `throw rejectOffer(` rather than `return rejectOffer(`. For example, with `return` a possible type checker might try to check the return type of `rejectOffer` against the return type of `bid`.

core/zoe/contracts/publicAuction.js
```diff
@@ -20,75 +20,111 @@ export const makeContract = harden((zoe, terms) => {
...
+        }
+
+        const [payoutRules] = zoe.getPayoutRuleMatrix(
+          harden([inviteHandle]),
+          terms.assays,
+        );
+
+        const ruleKinds = ['want', 'offer'];
+        if (!hasValidPayoutRules(ruleKinds, terms.assays, payoutRules)) {
+          return rejectOffer(zoe, terms.assays, inviteHandle);
```
erights: No error message?

core/zoe/contracts/publicAuction.js
```diff
@@ -20,75 +20,111 @@ export const makeContract = harden((zoe, terms) => {
...
+        }
 
-      const ruleKinds = ['offerExactly', 'wantAtLeast'];
+        if (
+          !isOverMinimumBid(
```
erights: "isOver" implies greater than. You should be checking that the bid is greater than or equal to the minimum bid. The name of this predicate should clearly reflect that.

In one of my previous sketches towards Zoe, I introduced a helper function `canTradeWith(payoutRules1, payoutRules2)` which just determined if a swap would preserve offer safety without deciding where the surplus should go. Enough contracts actually only to pairwise swap reallocations that it often proved useful, even though it doesn't cover all cases.

erights: Where do you check compatibility on the item? IOW, that the item the seller offers includes the item the bidder wants?

`canTradeWith` would cover both.

core/zoe/zoe.js
```diff
@@ -0,0 +1,380 @@
...
+  payoutRules.map(payoutRule => payoutRule.units.label.assay);
+
+/**
+ * Create an instance of Zoe.
+ *
+ * @param additionalEndowments pure or pure-ish endowments to add to evaluator
+ */
+const makeZoe = (additionalEndowments = {}) => {
+  // Zoe has two mints: a mint for invites and a mint for
+  // escrowReceipts. The invite mint can be used by a smart contract
```
erights: Comment stale. No more receipts.

core/zoe/zoe.js
```diff
@@ -0,0 +1,380 @@
...
+          exitRule: offerRules.exitRule,
+          assays,
+          payoutPromise: makePromise(),
+          units,
+        };
+        return offerTable.create(offerHandle, offerImmutableRecord);
+      },
+      /**
+       *  The contract can also create a real offer and get the
+       *  associated offerHandle, bypassing the escrow receipt
```
erights: Stale comment

core/zoe/zoe.js
```diff
@@ -0,0 +1,380 @@
...
+    });
+    return contractFacet;
+  };
+
+  // The public Zoe service has four main methods: `install` takes
+  // contract code and registers it with Zoe associated with an
+  // `installationHandle` for identification, `makeInstance` creates
+  // an instance from an installation, `getInstance` credibly
+  // retrieves an instance from zoe, and `escrow` allows users to
+  // securely escrow and get an escrow receipt and payouts in return.
```
erights: stale comment

core/zoe/zoe.js
```diff
@@ -0,0 +1,380 @@
...
+        payoutRules: offerRules.payoutRules,
+        exitRule: offerRules.exitRule,
+        assays,
+        payoutPromise: makePromise(),
+      };
+
+      // units should only be gotten after the payments are deposited
+      offerTable.create(offerHandle, offerImmutableRecord);
+
+      // Promise flow = assay -> purse -> deposit payment -> escrow receipt
```
erights: stale comment

erights (commented):
Still reviewing. Publishing what I have so far because we're about to meet.

core/zoe/contracts/coveredCall.js
katelynsills: yup, that's correct. So far the covered call, autoswap, and myFirstDapp have not been changed

core/zoe/contracts/publicAuction.js
katelynsills: there's a default error message

core/zoe/contracts/automaticRefund.js
katelynsills: there is an initial invite, but it's made by Zoe as part of the contract instantiation step, based on this  `initialInviteHandle` and the `initialSeat`. 

Let me see what I can do about making this simpler

core/zoe/contracts/helpers/userFlow.js
katelynsills: I've been trying to get away from having a canonical set of assays per instance, so I changed the parts of Zoe that assume that. It simplified state.js and the zoe code, although it does make it more complex for smart contracts that work fine with a canonical set of assays per instance. Is there a good way to get out of this tradeoff?

core/zoe/contracts/publicAuction.js
katelynsills: Good idea

erights:
> I've been trying to get away from having a canonical set of assays per instance, so I changed the parts of Zoe that assume that. It simplified state.js and the zoe code, although it does make it more complex for smart contracts that work fine with a canonical set of assays per instance. Is there a good way to get out of this tradeoff?

I didn't know that this was the purpose. Please proceed as is then, for now, and we can revisit later.

core/zoe/contracts/automaticRefund.js
```diff
@@ -11,18 +11,23 @@ import harden from '@agoric/harden';
  * @param {contractFacet} zoe - the contract facet of zoe
```
erights: stale mention of receipt above.

erights (commented):
Just a bit more review. But stopping because I am no longer absorbing. Probably cannot get back to things until Tuesday.

I really like the new automaticRefund!

core/zoe/contracts/helpers/offerRules.js
```diff
@@ -3,22 +3,22 @@ import { sameStructure } from '../../../../util/sameStructure';
...
   zoe,
+  assays,
   leftPayoutRules,
   rightPayoutRules,
 ) => {
   // "exactly matching" means that units are the same, but that the
   // kinds have switched places in the array
-  const extentOpsArray = zoe.getExtentOpsArray();
-  const exactlyKinds = ['wantExactly', 'offerExactly'];
+  const unitOpsArray = zoe.getUnitOpsForAssays(assays);
```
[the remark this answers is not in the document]
erights: oops. explained. resolving.

core/zoe/contracts/helpers/offerRules.js
```diff
@@ -62,47 +57,29 @@ const hasAssays = (assays, newPayoutRules) =>
...
- * @param {object} label - the label for the assay to use in the units
- * @param {*} allegedExtent - the extent to use in the units
- */
-export const makeUnits = (extentOps, label, allegedExtent) => {
-  extentOps.insistKind(allegedExtent);
-  return harden({
-    label,
-    extent: allegedExtent,
+    inviteHandles: active,
+    payoutRuleMatrix: zoe.getPayoutRuleMatrix(active, assays),
```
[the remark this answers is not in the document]
erights: oops. explained. resolving.

core/zoe/contracts/automaticRefund.js
```diff
@@ -20,14 +20,16 @@ export const makeContract = harden((zoe, terms) => {
...
       },
-      makeInvite: makeSeatInvite,
-      getOffersCount: () => offersCount,
     });
     const { invite, inviteHandle } = zoe.makeInvite(seat);
     return invite;
   };
   return harden({
     invite: makeSeatInvite(),
+    publicAPI: {
```
erights: These are supposed to be query-only right? (By convention that is, since we do not (currently) have any way of enforcing that.) I would be disturbed if these were understood as a normal way for a contract to expose effect-causing methods, so let's document that should only be used for query ops, and only do so ourselves. By that criteria, this one is fine. Although perhaps we want to adopt a name instead of `publicAPI` that emphasizes that.

katelynsills: I'm trying out something new, in that the publicAPI can be used to query and makeInvites if the invites aren't meant to be restricted. 

Is there a reason that you don't want to expose effect-causing methods, such as creating invites?

erights: Sorry, query-only was too narrow. Query-only + making invites is ok, where the *making* of the invite doesn't cause any interesting change of state by itself, and the capacity to change state can only be realized by redeeming the invite.

Not sure I can state the intuition, and not sure it's well motivated. But going beyond those bounds makes me itch. To be revisited later when I'm more focused.

katelynsills: Ok, glad to clear that up. I think I am only planning to do query-only + making invites. If anything outside of that seems necessary, let's revisit that

core/zoe/contracts/publicAuction.js
katelynsills: Sure, that's a good idea. Do you know where `canTradeWith` can be found? I can reproduce it if not

erights: I could try to find it. But it was written for a system sufficiently different that you'll probably rewrite it from scratch anyway ;)

erights:
> Yup, good point. I refactored. Let me know if that doesn't solve it

Fixed.

katelynsills:
Closed now that the content has been ported over to the monorepo

PR #232, opened 2019-11-21
package.json: roll back .dev0 version

warner:
This prepares for a monorepo, where we want packages/ERTP to provide
`0.1.12` (not `0.1.13-dev.0`), to be compatible with e.g.
packages/cosmic-swingset (which will declare a dependency upon `~0.1.12`,
which is not satisfied by `0.1.13-dev.0`).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@agoric/ertp",
-  "version": "0.1.13-dev.0",
+  "version": "0.1.12",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@agoric/ertp",
-  "version": "0.1.13-dev.0",
+  "version": "0.1.12",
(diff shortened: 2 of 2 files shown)


warner:
Things are going to be in flux as we get the monorepo set up, but the longer-term change (which we should totally discuss) is to stop the practice of bumping the `package.json` 's `version` field to `(N+1)-dev.0` just after releasing version `N`. In the monorepo, we need all the subdirectories to be compatible with each other (and we'll be using separate versions for each subdirectory). When the downstream package (e.g. cosmic-swingset) says it requires `~0.1.12`, then the sibling `ERTP` directory had better self-identify as `0.1.12`, or else the setup tools (`yarn install`) will correctly determine that it cannot use the `ERTP` directory to satisfy that dependency, and will install a new copy from NPM instead, defeating the purpose of the monorepo (which is to let us edit `ERTP/*.js` in place, and using those changes when running the cosmic-swingset tests).

We'll have to update our release-making process to avoid making these extra changes.. maybe it just means *not* running `npm version prerelease` after doing the `npm publish`.

PR #235, opened 2019-11-22
fix(zoe): have install accept a moduleFormat argument

michaelfig:
This is needed to be compatible with future bundle formats.

It apparently looks like this code can be simplified or eliminated, but it exists to explicitly declare what moduleFormats are accepted by Zoe.  When more formats are added or support for old ones is removed, this will be an important safeguard.

### DIFF
--- a/test/swingsetTests/zoe/test-zoe.js
+++ b/test/swingsetTests/zoe/test-zoe.js
@@ -108,7 +108,7 @@ test('zoe - coveredCall - valid inputs - with SES', async t => {
   }
 });
 
-test.only('zoe - coveredCall - valid inputs - no SES', async t => {
+test('zoe - coveredCall - valid inputs - no SES', async t => {
   try {
     const startingExtents = [[3, 0], [0, 7]];
     const dump = await main(false, 'zoe', [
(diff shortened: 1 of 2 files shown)


katelynsills:
Good idea. I had eliminated the switch statement since I had no plans to support other formats and it didn't seem likely to be confused about the format of Zoe contracts.  Can you say more about what you see the future formats as being?

test/swingsetTests/zoe/test-zoe.js
```diff
@@ -108,7 +108,7 @@ test('zoe - coveredCall - valid inputs - with SES', async t => {
   }
 });
 
-test.only('zoe - coveredCall - valid inputs - no SES', async t => {
```
Chris-Hibbert: Sorry

michaelfig:
> Can you say more about what you see the future formats as being?

Least authority module linkage.  The current `getExport` cannot support that mode as it requires more infrastructure on the server side.

katelynsills:
> Least authority module linkage. The current getExport cannot support that mode as it requires more infrastructure on the server side.

Ok, that is good to know. Do you see there being a change in how Zoe contracts are written to support this, or do you think all of the changes are in the `bundleSource` call?

michaelfig:
I think all the changes will be in `bundleSource`.

katelynsills:
Ok, thanks! This looks good to me then. I think in the future I will change `install` to take in the result of bundleSource such that the smart contract installer doesn't have to know/care about moduleFormats.

